PR #2, opened 2017-02-14
Support monitoring ports declared in kubernetes annotations.

stephen-soltesz:
Using the previous configuration, Prometheus would try to use port 80 on discovered pod addresses, but we need the flexibility to monitor alternate ports from (possibly) public services.

This PR adds support for monitoring ports declared in kubernetes annotations. For example:
    prometheus.io/port=9090

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/prometheus-support/2)


pboothe:
LGTM

PR #3, opened 2017-02-14
Correct regex for address and port configuration.

stephen-soltesz:
This change updates the regex to match addresses with or without default
ports for pod and service discovery.

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/prometheus-support/3)


pboothe:
L"G"TM

Some people, when confronted with a problem, think "I know, I'll use regular expressions." Now they have two problems. -jwz

PR #4, opened 2017-02-16
Fix kubernetes host:port relabel regex.

stephen-soltesz:
This change corrects a bug introduced by PR: https://github.com/m-lab/prometheus-support/pull/3

The regex uses three groups: the hostname, an optional port, and the
preferred port from a kubernetes annotation.

Previously, the second group should have been ignored if a :port was not
present in the input. However, making the port group optional with the
"?" had the unintended side-effect of allowing the hostname regex "(.+)"
to match greedily, which included the ":port" patterns up to the ";"
separating the hostname from the kubernetes port annotation.

This change updates the regex for the hostname to match any non-":"
characters. This forces the regex to stop if a ":port" is present and
allow the second group to match the optional port.

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/prometheus-support/4)

### DIFF
--- a/prometheus.yml
+++ b/prometheus.yml
@@ -149,6 +149,8 @@ scrape_configs:
         action: replace
         target_label: __address__
         # A google/re2 regex, matching addresses with or without default ports.
+        # NB: this will not work with IPv6 addresses. But, atm, kubernetes uses
+        # IPv4 addresses for internal network and GCE doesn not support IPv6.
         regex: ([^:]+)(?::\d+)?;(\d+)
         replacement: $1:$2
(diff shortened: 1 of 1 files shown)


pboothe:
This will break on v6 addresses. Is that okay? If so, LGTM

stephen-soltesz:
Yes. This will break with IPv6 addresses. At the moment, kubernetes on GCE uses IPv4 addresses for the internal container network, and GCE does not fully support IPv6.

We can revisit this regex when IPv6 becomes an issue because the regex will become considerably more complex to support literal v6 addresses.

PR #8, opened 2017-03-27
Add support for recording rules

stephen-soltesz:
This change adds recording rules to Prometheus.

Recording rules are a form of pre-calculation for expressions that are more expensive to execute at run-time. When active, recording rules can be queried like any other prometheus metric.

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/prometheus-support/8)


salarcon215:
Tested and works great. LGTM.

PR #10, opened 2017-03-27
Add disk usage monitoring of /prometheus mount point.

stephen-soltesz:
By default, (afaict) there is no monitoring of disk utilization for mount points of persistent volumes that uses the natural name familiar to operators, e.g. "/prometheus".

So, instead of a label like:
```
{mountpoint="/prometheus"}
```

Default behavior provides inscrutable labels like:
```
{mountpoint="/rootfs/var/lib/kubelet/pods/6950c80b-0b72-11e7-b1f8-42010af0004b/volumes/kubernetes.io~gce-pd/pvc-3e682a5b-f95d-11e6-a30a-42010af0004b"}
```

While the generated name does not change over time, it is a tedious manual process for an operator to translate that label to a human readable one.

This change provides human readable filesystem usage metrics by running a node-exporter within the same pod as the prometheus server. As a result, it has access to the same namespace and mount points as the prometheus-server.

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/prometheus-support/10)

### DIFF
--- a/k8s/prometheus.yml
+++ b/k8s/prometheus.yml
@@ -21,6 +21,11 @@ spec:
         # Note: run=prometheus-server should match a service config with a
         # public IP and port so that it is publically accessible.
         run: prometheus-server
+      annotations:
+        # Tell prometheus service discovery to scrape the node-exporter running
+        # within the prometheus-server pod.
+        prometheus.io/scrape: 'true'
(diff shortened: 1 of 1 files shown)


pboothe:
LGTM

PR #11, opened 2017-03-27
Add a DaemonSet for the node-exporter

stephen-soltesz:
This change adds a DaemonSet configuration for the node-exporter to run on every node in a cluster.

The node-exporter collects [a variety of system metrics][exporter] that we can use to assess whole-cluster health.

[exporter]: https://github.com/prometheus/node_exporter#enabled-by-default

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/prometheus-support/11)

### DIFF
--- a/README.md
+++ b/README.md
@@ -176,6 +176,12 @@ with the label "run=prometheus-server":
 
     kubectl create -f k8s/mlab-sandbox/<cluser-name>/services.yml
 
+Create the node-exporter daemonset. A [DaemonSet][daemonset] ensures that all
+nodes in the cluster run a copy of a pod. Our daemonset will run one instance of
+the node-exporter on every node.
+
+    kubectl create -f k8s/node-exporter-daemonset.yml
+
 Create the prometheus deployment. This step starts the actual prometheus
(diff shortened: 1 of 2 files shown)


stephen-soltesz:
Good question.

It turns out that cadvisor is bundled as part of the kubelet running on every node in a gke cluster. And, the "kubernetes-nodes" targets discovered by prometheus are the "cAdvisor" interfaces provided by the kubelet.

Before you asked, I didn't actually know the above. I just knew that there were not as many metrics exported by the "kubernetes-nodes" targets (now cadvisor) as by the node-exporter. ~133 vs ~474 unique metrics (ignoring labels) by default.

This PR was to provide access to the additional metrics. Now, we can actually say that the node-exporter is "in addition to" cAdvisor not "instead of."

While there may be some redundant information between the two services, I'm currently erring on the side of "collect everything and sort it out later" because a) there's nothing worse than not having a metric to inspect during an incident and b) I think the cost/benefit of assessing the value of each metric ahead of time is too low (or high?) -- has a poor return.

pboothe:
LGTM

PR #14, opened 2017-04-05
Add basic alertmanager configuration

stephen-soltesz:
This change adds the minimal alertmanager configuration and integration with the existing federated prometheus server configuration.

To establish the framework, this change includes one initial alert: are federated clusters offline?

Alerts are defined and evaluated by prometheus: `config/federation/prometheus/alerts.yml`

Alerts are received from prometheus, aggregated, and routed to the correct "receivers" by alertmanager: `config/federation/alertmanager/config.yml.template`

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/prometheus-support/14)


README.md
```diff
@@ -35,22 +35,13 @@ Many services like prometheus provide canonical docker images published to
 changing the configuration at run time using ConfigMaps. For detailed background
 see the [official docs][configmaps].
 
-Create the ConfigMaps for prometheus:
+Create the ConfigMap for prometheus (only create one for cluster or federation):
```
pboothe: "only create a ConfigMap for one of cluster or federation, not both"

stephen-soltesz: Done.

stephen-soltesz:
Regarding easier management of .yml files, there is an official tool [helm](https://github.com/kubernetes/helm) that uses yml templates (aka [charts](https://github.com/kubernetes/charts)) that allow you to populate some of these values from outside the yaml files.

When I was originally looking and learning, my thought was that it was too early to use and learn two layers of indirection. I also suspect there will be a different set of trade-offs using that tool. And, I suspect that we could find a way to define our own templates to reduce some of the volume of our current configs. But, I'm still not sure we've reached the point where it will be worth while to invest the time to try.

All that to say, I feel the desire for better management of .yml files too.

PR #16, opened 2017-04-08
Cleanup service.yml files for status-mlab-staging deployments

salarcon215:
- Removes obsolete `k8s/mlab-staging/services.yml`

- Renames directory `k8s/mlab-staging/status-mlab-staging` to `k8s/mlab-staging/prometheus-federation`

- Adds services.yml to  `k8s/mlab-staging/prometheus-federation`. This version includes specs for Prometheus (port 9090), Grafana (port 3000), Alertmanager (port 9093), and Blackbox exporter (port 9115).

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/prometheus-support/16)


stephen-soltesz:
LGTM.

Unfortunately, managing the content of the services files adds some overhead. It might be helpful (at the expense of a little more overhead) to separate each service definition into it's own file named after the service it contains, just like the deployment configs, one per deployment. This would make omissions a little easier to spot with a directory listing.

The consolation is that we don't add new services all that often. So, just leaving this thought open.

salarcon215:
That sounds like a good idea, and one that would make it easier to maintain tidy bundles of files per deployment. My preference would be to couple that with the addition of super-simple wrapper scripts to recreate deployments and their dependencies in one step. You mentioned the k8s devs might be working on quick deployment-bouncing tools; If the timing on that doesn't look imminent, ops would be helped by such wrappers.

PR #17, opened 2017-04-18
Add basic pushgateway support.

stephen-soltesz:
This change adds a new service definition for the push gateway, a k8s
config for the pushgateway deployment, and an update to the prometheus server
scrape configuration to collect metrics from the push gateway.

The pushgateway does not need persistent storage with the minimal configuration and it's not clear that we want that ultimately for our use cases anyway.

The configuration changes to the prometheus.yml scrape configuration can be deployed safely without the pushgatway deployment.

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/prometheus-support/17)

### DIFF
--- a/config/federation/prometheus/prometheus.yml
+++ b/config/federation/prometheus/prometheus.yml
@@ -318,3 +318,25 @@ scrape_configs:
         regex: .*
         target_label: __address__
         replacement: blackbox-public-service.default.svc.cluster.local:9115
+
+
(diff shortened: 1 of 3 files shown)


nkinkade:
LGTM, Stephen. With the caveat that my review is something of a rubber stamp, since I'm still unfamiliar with Prometheus configuration. The YAML syntax looks fine, and the values _seem_ reasonable without having any knowledge of the current system.

PR #18, opened 2017-04-18
Separate staging services

stephen-soltesz:
This change removes the combined service definition file `services.yml` in favor of one file per service definition.

This change is to make it more explicit when new services are added or when service configurations need to be updated.

In our configuration, all services share the same public IP address.

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/prometheus-support/18)


nkinkade:
LGTM

PR #22, opened 2017-05-03
Increase resources for federated prometheus

stephen-soltesz:
This change increases the resources assigned to Prometheus based on load requirements from work on the ETL pipeline.

As well, this provides a mount point for saving the aeflex service discovery file (accidentally omitted from a previous PR).

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/prometheus-support/22)

### DIFF
--- a/k8s/federation/prometheus.yml
+++ b/k8s/federation/prometheus.yml
@@ -50,11 +50,11 @@ spec:
           - containerPort: 9090
         resources:
           requests:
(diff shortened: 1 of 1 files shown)


nkinkade:
LGTM. I'm interested to see that they use the "Gi" label. I don't see that all that often in connection with RAM.

stephen-soltesz:
Thank you!

PR #23, opened 2017-05-08
Use ephemeral IP for prometheus cluster service configuration

stephen-soltesz:
This change updates the k8s/cluster configuration for prometheus so that the service definition uses an ephemeral IP that can be discovered later by searching for services annotated with the value: "gke-prometheus-federation/scrape: true".

This simplifies cluster deployments by not requiring a static ip allocation before hand and provides a means of discovering the target for "federation" scraping at a later time (likely from another cluster).


nkinkade:
LGTM

PR #25, opened 2017-05-08
Add blackbox exporter module for rsyncd.

stephen-soltesz:
This change adds a new module to the blackbox exporter configuration for checking the status of rsyncd servers.

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/prometheus-support/25)

### DIFF
--- a/config/federation/blackbox/config.yml
+++ b/config/federation/blackbox/config.yml
@@ -37,6 +37,17 @@ modules:
       query_response:
         - expect: "SSH-2.0-OpenSSH_.+"
 
+  # target=<hostname:port>
+  rsync_online:
+    prober: tcp
+    timeout: 10s
+    tcp:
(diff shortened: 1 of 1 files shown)


nkinkade:
LGTM. One question: do you not need to tell it which TCP port to probe?

stephen-soltesz:
Good question. The caller must specify the port. The hostname and port are defined through the `target=` parameter when making the blackbox request. See this example: https://github.com/m-lab/prometheus-support/blob/master/config/federation/blackbox/config.yml#L13

My next PR will be a change to mlabconfig to generate a blackbox targets file for this new rsync module, like this (but with the rsync modules/ports that scraper uses), https://github.com/m-lab/prometheus-support/blob/master/README.md#blackbox-exporter

PR #27, opened 2017-05-11
Fix blackbox timeout and prometheus relabel regex for blackbox module label

stephen-soltesz:
This change includes two fixes to the blackbox configuration.

First, we change all blackbox module timeouts to 5 seconds to guarantee that this timeout is less than the prometheus scrape timeout of 10 seconds. This is necessary to ensure that prometheus receives metrics about the failed probe rather than concluding that the blackbox exporter is down.

Second, we change a regex in the blackbox exporter relabel config to use parentheses so that the group numbers (`${1}`) work as intended. Their original omission was a typo.

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/prometheus-support/27)

### DIFF
--- a/config/federation/blackbox/config.yml
+++ b/config/federation/blackbox/config.yml
@@ -7,6 +7,11 @@
 # synchronously with the request. Typically, Prometheus will issue the request
 # to the blackbox exporter.
 #
+# The prober timeout setting must be less than the prometheus scrape timeout
+# or the prometheus server will give up and conclude that the blackbox service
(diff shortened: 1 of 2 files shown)


gfr10598:
:lgtm:

---

Review status: 0 of 2 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/m-lab/prometheus-support/27#-:-Kjtj5cLIx-eEnD8msDz:bnfp4nl)*

PR #31, opened 2017-05-22
Add service definitions for mlab-oti & deploy scripts

stephen-soltesz:
This change adds k8s service definitions for the mlab-oti deployment of the prometheus-federation cluster.

As well, we add two `apply-*` scripts that will make it easier to deploy changes in the future.


critzo:
LGTM thanks.

PR #33, opened 2017-05-23
Add nodeSelector for prometheus

stephen-soltesz:
To better support deployment to scraper, we need to add nodeSelectors so that prometheus only deploys to specific machines (and so scraper instances only deploy to different machines).

As well, this change includes commands for creating a cluster or adding a new node-pool for prometheus.

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/prometheus-support/33)


critzo:
LGTM @stephen-soltesz

PR #34, opened 2017-05-23
Allow longer timeouts for collection from federated clusters.

stephen-soltesz:
The scrape interval is every 60seconds and the default timeout is 10seconds.

This change increases the scrape timeout to 50s only for federation collection. This is not because collection *should* take that long or even that we want to allow it to take that long, but so that we have more headroom to monitor deployments with longer collection times than the default (10s).

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/prometheus-support/34)


critzo:
LGTM @stephen-soltesz

PR #40, opened 2017-05-24
Change pod service discovery to include all explicitly declared containerPorts

stephen-soltesz:
This change updates the pod service discovery behavior to use explicitly declared container ports as a way of identifying targets.

Also, this change updates the labels assigned to discovered k8s pod targets. In particular, the following labels are now assigned to pods:

 * container - the container name.
 * pod - the pod name.
 * deployment - the deployment name (derived from pod name)
 * namespace - the k8s namespace.
 * cluster - the k8s cluster and node-pool name.
 * ready - is the pod ready

TODO: if this works as intended, similar changes are needed to the `config/federation/prometheus` config.

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/prometheus-support/40)

### DIFF
--- a/config/cluster/prometheus/prometheus.yml
+++ b/config/cluster/prometheus/prometheus.yml
@@ -137,13 +137,19 @@ scrape_configs:
       - role: pod
 
     relabel_configs:
+      # If the pod is ready, keep it. Otierwise, ignore it.
+      - source_labels: [__meta_kubernetes_pod_ready]
+        regex: true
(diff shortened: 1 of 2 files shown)


stephen-soltesz:
Thank you for the careful review.

PTAL?

The `deployment` regex is still two steps but it is more correct and simpler than before. Attempts to combine the two regexes continued to fail.

PR #43, opened 2017-05-30
Use latest Prometheus version with more RAM and SSD

stephen-soltesz:
To accommodate the load generated by the scraper deployment in mlab-oti we're increasing the RAM used as well as using an SSD for metric storage.

This configuration has run over the weekend with 75% of the scraper instances. It is expected to remain stable at 100%.

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/prometheus-support/43)

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,13 +37,15 @@ gcloud container \
 
 No special scopes are required for the prometheus cluster configuration.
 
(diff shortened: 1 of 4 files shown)


critzo:
LGTM @stephen-soltesz

PR #66, opened 2017-07-05
Add the beginnings of a config for Annotation Service

jlm123hi:


config/federation/prometheus/prometheus.yml
```diff
@@ -415,3 +415,10 @@ scrape_configs:
       # Copy all labels from the aeflex service discovery.
       - action: labelmap
         regex: __aef_(.+)
+        
+  # Scrape config for the annotation service
+  - job_name: 'annotation-service'
+    static_configs:
+      - targets: ['annotator-dot-mlab-sandbox.appspot.com']
```
gfr10598: I think you will want to add the instances.  For now add a # TODO(jm) - Add individual instances once we figure out how.

jlm123hi: You're absolutely right, I'll add that.

PR #68, opened 2017-07-06
Sandbox marques

jlm123hi:
Trying to work around cloud networking issues.

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/prometheus-support/68)


k8s/mlab-sandbox/downloader/deployments/prometheus.yml
```diff
@@ -0,0 +1,145 @@
...
+        #   -storage.local.checkpoint-interval 10m0s
+        #
+        # Just use 10x the default values for all index cache flags:
+        #   -storage.local.index-cache-size.fingerprint-to-metric 104857600
+        #   -storage.local.index-cache-size.fingerprint-to-timerange 52428800
+        #   -storage.local.index-cache-size.label-name-to-label-values 104857600
+        #   -storage.local.index-cache-size.label-pair-to-fingerprints 20971520
+        args: ["-config.file=/etc/prometheus/prometheus.yml",
+               "-storage.local.path=/prometheus",
+               "-storage.local.target-heap-size=75161927680",
```
stephen-soltesz: Following up from our in person discussion, I believe it will be helpful to reduce or remove the target-heap-size flag and to reduce the "memory" & "cpu" request to a more proportionate size for a stand-alone cluster like the "downloader" cluster.

jlm123hi: Yeah, as expected, you are totally 100% right about this.

stephen-soltesz:
@gfr10598 some of this is obsolete -- and the base is very out of sync with the prometheus-support HEAD. I suggest identifying anything that we need to keep and creating a separate PR to preserve it.

@gfr10598 what needs to be preserved? The annotator is an AEFlex service, correct? (So, monitoring there should be automatic if the annotator has the right aeflex configuration). And, the downloader is monitored via a separate cluster.

Can this PR be closed?

stephen-soltesz:
Obsolete.

PR #79, opened 2017-08-04
Add scrape annotations for blackbox deployment

stephen-soltesz:
This PR adds the `prometheus.io/scrape: 'true'` to the blackbox service, previously missing. This will allow the prometheus server to scrape metrics about the blackbox service itself.

As well, this PR includes minor comment updates to the prometheus deployment configs, and enables the 'prometheus' service account in mlab-oti (a step that should have been completed already).

### DIFF
--- a/k8s/mlab-oti/prometheus-federation/deployments/blackbox.yml
+++ b/k8s/mlab-oti/prometheus-federation/deployments/blackbox.yml
@@ -21,6 +21,9 @@ spec:
         # Note: run=blackbox-server should match a service config with a
         # public IP and port so that it is publically accessible.
         run: blackbox-server
+      annotations:
+        # Tell prometheus service discovery to scrape the blackbox container.
+        prometheus.io/scrape: 'true'
     spec:
       # Place the pod into the Guaranteed QoS by setting equal resource
       # requests and limits for *all* containers in the pod.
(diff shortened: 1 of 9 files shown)


stephen-soltesz:
@nkinkade I discovered that the blackbox config was missing the `prometheus.io/scrape` annotation when I saw that the snmp-exporter wasn't being monitored directly in your new cluster.

I recommend also adding the following to your snmp-exporter.yml config:
```
      annotations:
        # Tell prometheus service discovery to scrape the snmp-exporter container.
        prometheus.io/scrape: 'true'
```

stephen-soltesz:
The annotations added here are interpreted by the prometheus service discovery configuration (which we'll talk through soon -- it will become very important for the snmp exporter soon).

The `prometheus.io/port` label is actually ignored now. So removing it was bringing this file into sync with what the service discovery config expects.

The current service discovery config uses the convention that if the annotation `prometheus.io/scrape: 'true'` is present, then scrape from every pod container that has an explicitly declared `containerPort`.

The situation is a little awkward because labels and annotations are not available on containers only at the pod level. So, this is an attempt to permit more than one container being monitored (impossible when we specified a single port) and give some control over which containers gets monitored. We may have to revisit this again if we need multiple ports from the same service where one should not be scraped. But we don't have that yet.

stephen-soltesz:
And, thank you for the review!

PR #86, opened 2017-09-28
Adds a block to configure the Prom cluster to scrape the snmp-targets.

nkinkade:
This change is [Reviewable](https://reviewable.io/reviews/m-lab/prometheus-support/86)

### DIFF
--- a/config/federation/prometheus/prometheus.yml
+++ b/config/federation/prometheus/prometheus.yml
@@ -402,7 +402,7 @@ scrape_configs:
         target_label: __param_module
         replacement: ${1}
 
-      # The "__exporter-project" label will contain a shortened GCP project
+      # The "__exporter_project" label will contain a shortened GCP project
       # name. We use this to construct the domain name of the GCE instance
       # running the snmp_exporer in that project.
       - source_labels: [__exporter_project]
--- a/config/federation/prometheus/prometheus.yml
+++ b/config/federation/prometheus/prometheus.yml
@@ -391,7 +391,7 @@ scrape_configs:
       # Here, we set (i.e. "replace") a request parameter "target" equal to the
       # host value.
       - source_labels: [__address__]
-        regex: (.*)(:9116)?
+        regex: (.*)
         target_label: __param_target
         replacement: ${1}
 
--- a/config/federation/prometheus/prometheus.yml
+++ b/config/federation/prometheus/prometheus.yml
@@ -405,7 +405,7 @@ scrape_configs:
       # The "__exporter-project" label will contain a shortened GCP project
       # name. We use this to construct the domain name of the GCE instance
       # running the snmp_exporer in that project.
-      - source_labels: [__exporter-project]
+      - source_labels: [__exporter_project]
         regex: (.*)
         target_label: __address__
         replacement: snmp-exporter.${1}.measurementlab.net:9116
--- a/config/federation/prometheus/prometheus.yml
+++ b/config/federation/prometheus/prometheus.yml
@@ -373,6 +373,42 @@ scrape_configs:
         target_label: __address__
         replacement: blackbox-public-service.default.svc.cluster.local:9115
 
+  # SNMP configurations.
+  - job_name: 'snmp-targets'
+    metrics_path: /snmp
+
+    file_sd_configs:
+      - files:
+          - /snmp-targets/*.json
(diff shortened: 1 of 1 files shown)


config/federation/prometheus/prometheus.yml
```diff
@@ -373,6 +373,42 @@ scrape_configs:
...
+      # the snmp_exporter request.
+      - source_labels: [site]
+        regex: (.*)
+        target_label: __param_module
+        replacement: ${1}
+
+      # The "__exporter-project" label will contain a shortened GCP project
+      # name. We use this to construct the domain name of the GCE instance
+      # running the snmp_exporer in that project.
+      - source_labels: [__exporter-project]
```
stephen-soltesz: Please let's use all underscores in label names, i.e. `__exporter_project`.

config/federation/prometheus/prometheus.yml
```diff
@@ -373,6 +373,42 @@ scrape_configs:
...
+
+    # This relabel config is necessary. The relabel config redefines the address
+    # to scrape and sets the correct parameters to pass to the scrape target.
+    relabel_configs:
+
+      # The default __address__ value is a target host from the config file.
+      # Here, we set (i.e. "replace") a request parameter "target" equal to the
+      # host value.
+      - source_labels: [__address__]
+        regex: (.*)(:9116)?
```
stephen-soltesz: Does this work in practice? My expectation is that it actually strips off the port suffix. The port 80 pattern from the blackbox exporter is b/c prometheus helpfully adds `:80` to target addresses that have no port already.

config/federation/prometheus/prometheus.yml
nkinkade: Fixed, both here and in m-lab/operator/generate_prometheus_targets.sh.

config/federation/prometheus/prometheus.yml
nkinkade: It works, I think, because the target generated by mlabconfig.py is currently using a template like [s1.{{site}}.measurement-lab.org:9116](https://github.com/m-lab/operator/blob/master/generate_prometheus_targets.sh#L47). But maybe the target shouldn't have the port added, and I can simply add it here?

config/federation/prometheus/prometheus.yml
```diff
@@ -373,6 +373,42 @@ scrape_configs:
...
+        replacement: ${1}
+
+      # Use the "site" label defined in the input file as the module name for
+      # the snmp_exporter request.
+      - source_labels: [site]
+        regex: (.*)
+        target_label: __param_module
+        replacement: ${1}
+
+      # The "__exporter-project" label will contain a shortened GCP project
```
stephen-soltesz: nit: and `__exporter_project` here. (sorry).

config/federation/prometheus/prometheus.yml
```diff
@@ -373,6 +373,42 @@ scrape_configs:
...
+
+    # This relabel config is necessary. The relabel config redefines the address
+    # to scrape and sets the correct parameters to pass to the scrape target.
+    relabel_configs:
+
+      # The default __address__ value is a target host from the config file.
+      # Here, we set (i.e. "replace") a request parameter "target" equal to the
+      # host value.
+      - source_labels: [__address__]
+        regex: (.*)
```
stephen-soltesz: Sorry. I was unclear. What I think will happen is that now prometheus will add `:80` to the end of the address and the value of `__param_target` will be the site switch hostname with port 80 at the end, e.g. `s1.lga1t.measurement-lab.org:80`. When I said "prometheus helpfully adds" the port I meant the "please don't help me" kind of helpful.  Though, this is from memory, so I could be mistaken.

nkinkade: I found [some random stackoverflow thread](https://stackoverflow.com/questions/40020053/how-to-relabel-address-with-its-current-value-and-a-label) that somewhat makes it *seem* like Prometheus won't automatically append `:80` to the `__address__` label when interprets `target` from the config. I guess I'll leave it like it is, and if it doesn't work then the fix will be easy.

config/federation/prometheus/prometheus.yml
nkinkade: Fixed.

PR #107, opened 2017-10-24
Drops all 'ifAlias' labels that are just empty strings. 

nkinkade:
In other words, only keep the label if it has a value. The vast majority of `ifAlias` labels will have no value. Don't pollute the datastore with empty labels, but instead just drop empty ones and only keep ones with some value.

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/prometheus-support/107)

### DIFF
--- a/config/federation/prometheus/prometheus.yml
+++ b/config/federation/prometheus/prometheus.yml
@@ -408,6 +408,13 @@ scrape_configs:
         target_label: __address__
         replacement: snmp-exporter.${1}.measurementlab.net:9116
 
+      # Most all 'ifAlias' labels will be empty. Rather than pollute the
+      # datastore with a bunch of empty labels, only keep the ones that have
+      # some value.
+      - source_labels: [ifAlias]
+        regex: .+
+        action: labelkeep
+
   # Scrape config for pushgateway.
   #
   # Push gateways allow services to publish metrics for later collection by
--- a/config/federation/prometheus/prometheus.yml
+++ b/config/federation/prometheus/prometheus.yml
@@ -416,8 +416,9 @@ scrape_configs:
       # datastore with a bunch of empty labels, only keep the ones that have
       # some value.
       - source_labels: [ifAlias]
-        regex: .+
-        action: labelkeep
+        regex: ^$
+        target_label: ifAlias
+        replacement: ""
 
   # Scrape config for pushgateway.
   #
--- a/config/federation/prometheus/prometheus.yml
+++ b/config/federation/prometheus/prometheus.yml
@@ -408,6 +408,10 @@ scrape_configs:
         target_label: __address__
         replacement: snmp-exporter.${1}.measurementlab.net:9116
 
+    # This relabel config is the last relabel processed before ingesting data
+    # into the datastore.
+    metric_relabel_configs:
+
       # Most all 'ifAlias' labels will be empty. Rather than pollute the
       # datastore with a bunch of empty labels, only keep the ones that have
       # some value.


stephen-soltesz:
Review status: 0 of 1 files reviewed at latest revision, 1 unresolved discussion.

---

*[config/federation/prometheus/prometheus.yml, line 411 at r1](https://reviewable.io:443/reviews/m-lab/prometheus-support/107#-KxFOzc7stRcb6y7oXNY:-KxFO3FgrVcBDxmz7z5B:b-gpeb07) ([raw file](https://github.com/m-lab/prometheus-support/blob/84d0388608edb78f97ca51b59d980f17cd5984f5/config/federation/prometheus/prometheus.yml#L411)):*
> ```YAML
>         replacement: snmp-exporter.${1}.measurementlab.net:9116
> 
>       # Most all 'ifAlias' labels will be empty. Rather than pollute the
> ```

Hmm. I've never used `labelkeep` before.

First thing, is I think we'll want this logic on a metric_relabel_configs section.  Since the ifAlias is defined on data scraped by prometheus.

Second thing, I'm not sure now if prometheus does what we want. The docstring for keeplabel says "Match regex against all label names. Any label that does not match will be removed from the set of labels."  That sounds like it's looking at the label name "ifAlias" rather than the value.

We can try this of course, the docs could be misleading.

But, if it does mean what it looks like, then it's not apparent which of the other options will do what we want either.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/m-lab/prometheus-support/107)*

nkinkade:
PTAL?

---

Review status: 0 of 1 files reviewed at latest revision, 1 unresolved discussion.

---

*[config/federation/prometheus/prometheus.yml, line 411 at r1](https://reviewable.io:443/reviews/m-lab/prometheus-support/107#-KxFOzc7stRcb6y7oXNY:-KxFRY9setSf7irXWrPn:bbo2ba0) ([raw file](https://github.com/m-lab/prometheus-support/blob/84d0388608edb78f97ca51b59d980f17cd5984f5/config/federation/prometheus/prometheus.yml#L411)):*

Hmm. I've never used `labelkeep` before.

First thing, is I think we'll want this logic on a metric_relabel_configs section.  Since the ifAlias is defined on data scraped by prometheus.

Second thing, I'm not sure now if prometheus does what we want. The docstring for keeplabel says "Match regex against all label names. Any label that does not match will be removed from the set of labels."  That sounds like it's looking at the label name "ifAlias" rather than the value.

We can try this of course, the docs could be misleading.

But, if it does mean what it looks like, then it's not apparent which of the other options will do what we want either.
</blockquote></details>

Thanks for catching my error on using `relabel_configs` instead of `metric_relabel_configs`. Fixed, I think.

The documentation you point out gave me pause as well. But after going over it a few times I concluded that it simply meant it would iterate over the 'label names' specified in the `source_labels` declaration, in this case being only `[ifAlias]`.  However, since you called it out, I dug into it more and found that your interpretation was correct:

https://github.com/prometheus/prometheus/issues/2367
https://github.com/prometheus/prometheus/issues/2368

Indeed, apparently `source_labels` and `labelkeep` are mutually exclusive.

The links above, plus another one I found seem to indicate you can remove a label by setting `replacement: ""`. I've modified the rules.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/m-lab/prometheus-support/107#-:-KxFWBAhMCvXiAtMRW6m:b-833305)*

stephen-soltesz:
:lgtm:

---

Review status: 0 of 1 files reviewed at latest revision, 1 unresolved discussion.

---

*[config/federation/prometheus/prometheus.yml, line 411 at r1](https://reviewable.io:443/reviews/m-lab/prometheus-support/107#-KxFOzc7stRcb6y7oXNY:-KxF_nN-rDtxOmMTtVAu:bjckrw2) ([raw file](https://github.com/m-lab/prometheus-support/blob/84d0388608edb78f97ca51b59d980f17cd5984f5/config/federation/prometheus/prometheus.yml#L411)):*

Thanks for catching my error on using `relabel_configs` instead of `metric_relabel_configs`. Fixed, I think.

The documentation you point out gave me pause as well. But after going over it a few times I concluded that it simply meant it would iterate over the 'label names' specified in the `source_labels` declaration, in this case being only `[ifAlias]`.  However, since you called it out, I dug into it more and found that your interpretation was correct:

https://github.com/prometheus/prometheus/issues/2367
https://github.com/prometheus/prometheus/issues/2368

Indeed, apparently `source_labels` and `labelkeep` are mutually exclusive.

The links above, plus another one I found seem to indicate you can remove a label by setting `replacement: ""`. I've modified the rules.
</blockquote></details>

Okay. Interesting. Thank you for doing the research. It's funny b/c on it's face that looks like a no-op -- replace the empty value with an empty value, but if that does the trick, that's great.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/m-lab/prometheus-support/107#-:-KxFaNRwxuXKmc099D-C:bnfp4nl)*

nkinkade:

---

*[config/federation/prometheus/prometheus.yml, line 411 at r1](https://reviewable.io:443/reviews/m-lab/prometheus-support/107#-KxFOzc7stRcb6y7oXNY:-KxFbt9o2dnIAKwgGRvD:b6qmw6q) ([raw file](https://github.com/m-lab/prometheus-support/blob/84d0388608edb78f97ca51b59d980f17cd5984f5/config/federation/prometheus/prometheus.yml#L411)):*

Okay. Interesting. Thank you for doing the research. It's funny b/c on it's face that looks like a no-op -- replace the empty value with an empty value, but if that does the trick, that's great.
</blockquote></details>

I don't know if that will do the trick, actually. I'm just going off of these:

https://groups.google.com/forum/#!msg/prometheus-developers/zLf13CKY8uY/aXAyvW7bBwAJ
https://github.com/prometheus/prometheus/issues/2367#issuecomment-275159798

These comments would make it appear that setting `replacement: ""` is some sort of undocumented feature that will actually just remove the entire label, and not actually just set the value to an empty string, like intuition would tell me. We'll find out soon, though. :)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/m-lab/prometheus-support/107)*

PR #116, opened 2017-11-02
Add slow-targets configuration

stephen-soltesz:
For the bigquery exporter it is helpful to poll the exporter less often than every minute.

This change adds a "slow-5m-targets" job configuration that scrapes targets every 5 min instead of the global 1 min.

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/prometheus-support/116)

### DIFF
--- a/config/federation/prometheus/prometheus.yml.template
+++ b/config/federation/prometheus/prometheus.yml.template
@@ -476,6 +476,14 @@ scrape_configs:
 
   # Scrape slow targets every 5m (NOTE: do not poll less frequently without
   # updating the prometheus flag -query.staleness-delta).
+  #
+  # TODO(soltesz): make this pattern more generic and consistent. Consider:
+  #  * per-target directories exclusively.
+  #  * create per-interval directories instead of per-target directories,
+  #      e.g. 1m-targets, 5m-targets, 30m-targets.
+  #  * place all target files into a single directory, named appropriately.
+  #      e.g. blackbox-1m-*.json, bq-exporter-5m-*.json, etc.
+  #
   - job_name: 'slow-5m-targets'
     scrape_interval: 5m
     file_sd_configs:
--- a/config/federation/prometheus/prometheus.yml.template
+++ b/config/federation/prometheus/prometheus.yml.template
@@ -473,3 +473,14 @@ scrape_configs:
       # Copy all labels from the aeflex service discovery.
       - action: labelmap
         regex: __aef_(.+)
+
+  # Scrape slow targets every 5m (NOTE: do not poll less frequently without
+  # updating the prometheus flag -query.staleness-delta).
+  - job_name: 'slow-5m-targets'
+    scrape_interval: 5m
+    file_sd_configs:
+      - files:
+          - /slow-targets/*-5m.json
+        # Attempt to re-read files every five minutes.
+        refresh_interval: 5m
+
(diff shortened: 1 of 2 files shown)


nkinkade:
The one thought I have is that this breaks our (unwritten?) convention of having all target configs from one tool/resource in their own intuitively named directory.  As snmp_exporter targets go in the snmp-exporter/ directory, and blackbox_exporter targets go in the blackbox-targets/ dir, federation-targets in the federation-targets/ dir, etc.  In this new case, we drop all configs that we want scraped at a 5m interval into a generic directory.  This isn't necessary a problem.

It occurs to me that we may want to consider one of a few things:

* To keep with existing convention, change the directory name from `slow-5m-targets` to `bigquery-exporter`, leaving all other configs, including`scraper_interval`, as it.

* Note: this may not even be possible because of related configs that are different between scrape targets (like `metrics_path`.  With that in mind, reorganize how we manage most scrapes by creating several jobs like `1m-targets`, `5m-targets`, `30m-targets`, and then dropping targets into the appropriate job bucket, such that we might have target locations like `1m-targets/snmpexporter.json` and `1m-targets/blackboxexporter.json`. ... basically, any targets in a *.json file in `1m-targets/` gets scraped at 1m intervals. I mention this because it seems possible that there could be a single tool/resource where we may not want to scrape every target at the same interval. For example, perhaps we want to scrape `ssh806` blackbox_exporter targets once a minute, but only really to scrape `rsyncd` blackbox_exporter targets once every 15 minutes. Right now we can't do that and keep our files well organized.

* We could just dump all target definition files into a single directory (e.g., `targets/`)  and name the files appropriately. A file named like `*-5m.json` gets automatically picked up by a job that scrapes at 5m intervals (which is the same as what you are doing in this PR), etc.

There are just some thoughts. If it's too much for now, then LGTM. :)

:lgtm: 

---

Reviewed 2 of 2 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/m-lab/prometheus-support/116#-:-Kxxk-6OlOwGRxTDbYR3:b-eqm72d)*

stephen-soltesz:
Your questions are good. I agree that the convention up to now has been to create directories for different kinds of targets. In part this was to setup per-target relabel rules (blackbox, snmp, and aeflex, in particular). But many of the others don't use relabel rules. And in particular the one-off configs for nagios exporter, snmp exporter itself, and bq exporter (planned) may be better in target config files that we can control separately from the prometheus config.

I've added a TODO with a summary of your observations / ideas.

Right now I am inclined to defer this design question to prevent making a bunch of changes so that we can get the bq exporter working at all.

I like your suggestion of using one directory with appropriate file names. For example it does seem plausible to use patterns on two levels. like:  `*-<tool>-<interval>.json` or maybe `<tool>-<interval>-*.json` so files sort together.

So, blackbox exporter pattern would be `blackbox-1m-*.json`. But if we needed a 5m set of blackbox targets, we'd create a 5m rule and create files with the pattern `blackbox-5m-*.json`. I can imagine that working.

---

Review status: 1 of 2 files reviewed at latest revision, 1 unresolved discussion, some commit checks pending.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/m-lab/prometheus-support/116#-:-KxxsKLRjEh12QIhWOB0:b-i2gvoc)*

PR #134, opened 2017-11-29
Add ParserDailyVolumeTooLow alert

stephen-soltesz:
This change adds a new alert for the M-Lab test parsers.

The alert checks whether today's test volume is less than 20% of yesterday's volume. Since the parser schedule for all test data is periodic this appears to work well for NDT, and paris traceroute parsers. The alert explicitly ignores "batch" jobs which do not necessarily follow a predictable schedule.

This change also includes a complementary change to ScraperMostRecentArchivedFileTimeIsTooOld by extending the threshold to 56hours. This provides enough time for daily configuration jobs to run. And, prevents creation of a redundant alert with ParserDailyVolumeTooLow when many machines are affected.

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/prometheus-support/134)

### DIFF
--- a/config/federation/prometheus/alerts.yml
+++ b/config/federation/prometheus/alerts.yml
@@ -249,10 +249,15 @@ ALERT SnmpExporterDownOrMissing
   }
 
 
-# ParserDailyVolumeTooLow: today's test volume has dropped over 80% compared to yesterday's volume. The alert condition ignores batch processing.
+# ParserDailyVolumeTooLow: today's test volume has dropped over 80% compared to yesterday's
+# or last week's test volume. The alert condition ignores batch processing.
 ALERT ParserDailyVolumeTooLow
   IF sum by(service) (increase(etl_test_count{service!~".*batch.*"}[2h]))
-        < 0.20 * sum by(service) (increase(etl_test_count[2h] offset 24h))
+     < 0.20 * max by(service) (
+         sum by(service) (increase(etl_test_count[2h] offset 24h))
+            OR
+         sum by(service) (increase(etl_test_count[2h] offset 1w))
+     )
   FOR 1h
   LABELS {
     severity = "page"
--- a/config/federation/prometheus/alerts.yml
+++ b/config/federation/prometheus/alerts.yml
@@ -254,7 +254,7 @@ ALERT SnmpExporterDownOrMissing
 # ignores batch processing.
 ALERT ParserDailyVolumeTooLow
   IF sum by(service) (increase(etl_test_count{service!~".*batch.*"}[2h]))
-     < 0.20 * quantile by(service) (.5,
+     < 0.20 * quantile by(service) (0.50,
          sum by(service) (increase(etl_test_count[2h] offset 1d))
             OR
          sum by(service) (increase(etl_test_count[2h] offset 3d))
--- a/config/federation/prometheus/alerts.yml
+++ b/config/federation/prometheus/alerts.yml
@@ -257,9 +257,11 @@ ALERT ParserDailyVolumeTooLow
      < 0.20 * quantile by(service) (.5,
          sum by(service) (increase(etl_test_count[2h] offset 1d))
             OR
-         sum by(service) (increase(etl_test_count[2h] offset 2d))
-            OR
          sum by(service) (increase(etl_test_count[2h] offset 3d))
+            OR
+         sum by(service) (increase(etl_test_count[2h] offset 5d))
+            OR
+         sum by(service) (increase(etl_test_count[2h] offset 7d))
      )
   FOR 1h
   LABELS {
--- a/config/federation/prometheus/alerts.yml
+++ b/config/federation/prometheus/alerts.yml
@@ -85,7 +85,7 @@ ALERT ScraperMostRecentArchivedFileTimeIsTooOld
   }
   ANNOTATIONS {
     summary = "Scraper max file mtime is too old {{ $labels.rsync_url }}",
-    description = "Max file mtime for {{ $labels.rsync_url }} is older than 60 hours.",
+    description = "Max file mtime for {{ $labels.rsync_url }} is older than 56 hours.",
   }
 
 # Scraper internal consistency.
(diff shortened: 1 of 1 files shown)


gfr10598:
:lgtm:
One suggestion for avoiding false positives.

---

Review status: 0 of 1 files reviewed at latest revision, 1 unresolved discussion.

---

*[config/federation/prometheus/alerts.yml, line 253 at r1](https://reviewable.io:443/reviews/m-lab/prometheus-support/134#-L-8d_ZCGigmz9QAxjzK:-L-8d_ZCGigmz9QAxjzL:brsq0lk) ([raw file](https://github.com/m-lab/prometheus-support/blob/6db13b84091ebd605678a9a11e226b60c68f0834/config/federation/prometheus/alerts.yml#L253)):*
> ```YAML
> 
> # ParserDailyVolumeTooLow: today's test volume has dropped over 80% compared to yesterday's volume. The alert condition ignores batch processing.
> ALERT ParserDailyVolumeTooLow
> ```

How about AND for  same period yesterday and a week ago?  That should avoid most false positives when there was an anomaly in the past that shouldn't be repeated "today".

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/m-lab/prometheus-support/134#-:-L-8dxkuxgKlMZr0mqbE:b-j5bxss)*

stephen-soltesz:
Thank you!

---

---

*[config/federation/prometheus/alerts.yml, line 253 at r1](https://reviewable.io:443/reviews/m-lab/prometheus-support/134#-L-8d_ZCGigmz9QAxjzK:-L-8jCwEd7kxGjzx60Kh:b-npkf10) ([raw file](https://github.com/m-lab/prometheus-support/blob/6db13b84091ebd605678a9a11e226b60c68f0834/config/federation/prometheus/alerts.yml#L253)):*

How about AND for  same period yesterday and a week ago?  That should avoid most false positives when there was an anomaly in the past that shouldn't be repeated "today".
</blockquote></details>

Ah, good idea.  Done.

The trick I used here is `OR` which combines the "yesterday" set with the "1week ago" set over which `max` operates. And because the LHS of the `<` uses `sum by(service) ()` and the RHS uses `max by(service) ()`, the LHS and RHS labels should match. I've left out the anti-batch pattern on the RHS b/c only the LHS values matter for the alert.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/m-lab/prometheus-support/134#-:-L-8jrWCpvJB4wlOwSB-:b-9755fn)*

stephen-soltesz:

---

*[config/federation/prometheus/alerts.yml, line 253 at r1](https://reviewable.io:443/reviews/m-lab/prometheus-support/134#-L-8d_ZCGigmz9QAxjzK:-L-8kEMpOhf966l7DJWq:bv60q01) ([raw file](https://github.com/m-lab/prometheus-support/blob/6db13b84091ebd605678a9a11e226b60c68f0834/config/federation/prometheus/alerts.yml#L253)):*

Ah, good idea.  Done.

The trick I used here is `OR` which combines the "yesterday" set with the "1week ago" set over which `max` operates. And because the LHS of the `<` uses `sum by(service) ()` and the RHS uses `max by(service) ()`, the LHS and RHS labels should match. I've left out the anti-batch pattern on the RHS b/c only the LHS values matter for the alert.
</blockquote></details>

I realized that `max` is not what we want. This would also make the recovery surge problematic in the future. Neither is `min` what we want to avoid setting the "down" period as the standard.

I think we want something like the `avg` or median value from the last several days. I've tried using this approach over the last three days.

How does this look?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/m-lab/prometheus-support/134)*

stephen-soltesz:
PTAL?

---

Review status: 0 of 1 files reviewed at latest revision, 1 unresolved discussion, some commit checks pending.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/m-lab/prometheus-support/134#-:-L-8o-jQTZg_wZNScfg5:b-833305)*

gfr10598:
:lgtm:

---

Review status: 0 of 1 files reviewed at latest revision, 2 unresolved discussions, all commit checks successful.

---

*[config/federation/prometheus/alerts.yml, line 262 at r2](https://reviewable.io:443/reviews/m-lab/prometheus-support/134#-L-8obVpffgTfqhXw7-W:-L-8obVpffgTfqhXw7-X:b5e6coz) ([raw file](https://github.com/m-lab/prometheus-support/blob/da588cb0802c69a642c13a46f94d443114557ae4/config/federation/prometheus/alerts.yml#L262)):*
> ```YAML
>          sum by(service) (increase(etl_test_count[2h] offset 2d))
>             OR
>          sum by(service) (increase(etl_test_count[2h] offset 3d))
> ```

I like it, but would suggest you use 1w instead of 3d.  Alternatively, use 1w, 2w, 3w for even greater robustness?  Or 1w,2w,3w,1d,2d ??

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/m-lab/prometheus-support/134#-:-L-8owwhTlMaummD9s8n:bnfp4nl)*

stephen-soltesz:
Review status: 0 of 1 files reviewed at latest revision, 2 unresolved discussions, some commit checks pending.

---

*[config/federation/prometheus/alerts.yml, line 262 at r2](https://reviewable.io:443/reviews/m-lab/prometheus-support/134#-L-8obVpffgTfqhXw7-W:-L-8s90u6UaNsyoYglgU:bgr5nzf) ([raw file](https://github.com/m-lab/prometheus-support/blob/da588cb0802c69a642c13a46f94d443114557ae4/config/federation/prometheus/alerts.yml#L262)):*

I like it, but would suggest you use 1w instead of 3d.  Alternatively, use 1w, 2w, 3w for even greater robustness?  Or 1w,2w,3w,1d,2d ??
</blockquote></details>

How about 1d, 3d, 5d, 7d? -- I'm concerned about periods longer than a week being constraining.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/m-lab/prometheus-support/134)*

stephen-soltesz:
Discussed offline and we'll try the 1, 3, 5, & 7d format.

PR #136, opened 2017-11-30
Fix ipv6 bias query & remove ndt_test_count_total.sql

stephen-soltesz:
This change adds the machine label to results in the ipv6 bias query. This is necessary to provide globally unique sets of labels on prometheus metrics. Using only 'site' was sufficient for the handful of sites before, but now we need per-machine labels.

This change also removes the expensive query for counting ndt tests. This was useful for highlighting the 1B test mark, but is too expensive a query to maintain in its current form.

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/prometheus-support/136)


gfr10598:
:lgtm:

---

Reviewed 3 of 3 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/m-lab/prometheus-support/136#-:-L-Dz_S4w7nqTagdu9on:bnfp4nl)*

Issue #144, opened 2017-12-04
Do not collect duplicate metrics 


stephen-soltesz:
The mlab-oti prometheus logs report that some metrics are duplicates. I'm not sure where this is coming from. Need to enable debug logging to get a sample by sample log of the sources.

```
 -log.level=debug
```

```
time="2017-12-04T18:00:09Z" level=warning msg="Error on ingesting samples with different value but same timestamp" numDropped=938 source="scrape.go:537" 
time="2017-12-04T18:01:21Z" level=warning msg="Error on ingesting samples with different value but same timestamp" numDropped=581 source="scrape.go:537" 
time="2017-12-04T18:02:10Z" level=warning msg="Error on ingesting samples with different value but same timestamp" numDropped=249 source="scrape.go:537"
```

PR #145, opened 2017-12-04
Add recording rules for NDT Early warnings

stephen-soltesz:
This change adds four new recording rules to the prometheus configuration for the NDT early warning system.

This change also enables debug logging on the Prometheus server to help locate the source of "dropped metric" errors cited in https://github.com/m-lab/prometheus-support/issues/144

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/prometheus-support/145)

### DIFF
--- a/config/federation/prometheus/rules.yml
+++ b/config/federation/prometheus/rules.yml
@@ -49,6 +49,7 @@ machine:inotify_extension_create:rpm2m =
     # NOTE: using 'without' instead of 'by' preserves all other labels.
     60.0 * sum without(ext) (rate(inotify_extension_create_total{ext=~".*_snaplog"}[2m]))
 
+# TODO: aggregate on per-machine aliases when available.
 # Per-switch "Out" (i.e. Download) bits per second.
 # Units: bits per second.
 uplink:ifHCOutOctets:bps2m =
--- a/config/federation/prometheus/rules.yml
+++ b/config/federation/prometheus/rules.yml
@@ -44,18 +44,22 @@ lsb:sidestream_connection_count:increase2m =
 ## NDT Early Warning aggregation rules.
 
 # Per-machine inotify creation rates, using only c2s_snaplog + s2c_snaplog files.
-machine:inotify_extension_create:rate2m =
+# Units: requests per minute.
+machine:inotify_extension_create:rpm2m =
     # NOTE: using 'without' instead of 'by' preserves all other labels.
     60.0 * sum without(ext) (rate(inotify_extension_create_total{ext=~".*_snaplog"}[2m]))
 
 # Per-switch "Out" (i.e. Download) bits per second.
-uplink:ifHCOutOctets:bits_per_second_rate2m =
+# Units: bits per second.
+uplink:ifHCOutOctets:bps2m =
     8 * rate(ifHCOutOctets{ifAlias="uplink"}[2m])
 
-# Ratio of time spent performing I/O on the /vservers partition.
-vserver_partition:node_disk_io_time_ms:ratio_rate2m =
-    rate(node_disk_io_time_ms{device="dm-2"}[2m]) / 1000
+# The maximum ratio of time spent performing I/O of all devices per machine.
+# Units: none
+machine:node_disk_io_time_ms:max_ratio2m =
+    max without(device) (rate(node_disk_io_time_ms{service="nodeexporter"}[2m])) / 1000
 
 # NDT vserver disk quota utilization, 12 hour estimate.
+# Units: KB
 ndt:vdlimit_used:predict_linear1h_12h =
     predict_linear(vdlimit_used{experiment="ndt.iupui"}[1h], 12*60*60)
--- a/config/federation/prometheus/rules.yml
+++ b/config/federation/prometheus/rules.yml
@@ -55,7 +55,7 @@ machine:inotify_extension_create:rpm2m =
 switch:ifHCOutOctets:bps2m =
     8 * rate(ifHCOutOctets{ifAlias="uplink"}[2m])
 
-# The maximum ratio of time spent performing I/O of all devices per machine.
+# Per-machine maximum ratio of time spent performing I/O on all devices.
 # Units: none
 machine:node_disk_io_time_ms:max_ratio2m =
     max without(device) (rate(node_disk_io_time_ms{service="nodeexporter"}[2m])) / 1000
--- a/config/federation/prometheus/rules.yml
+++ b/config/federation/prometheus/rules.yml
@@ -49,7 +49,7 @@ machine:inotify_extension_create:rpm2m =
     # NOTE: using 'without' instead of 'by' preserves all other labels.
     60.0 * sum without(ext) (rate(inotify_extension_create_total{ext=~".*_snaplog"}[2m]))
 
-# TODO: aggregate on per-machine aliases when available.
+# TODO: aggregate on per-machine interface aliases when available.
 # Per-switch "Out" (i.e. Download) bits per second.
 # Units: bits per second.
 switch:ifHCOutOctets:bps2m =
--- a/config/federation/prometheus/rules.yml
+++ b/config/federation/prometheus/rules.yml
@@ -6,15 +6,20 @@
 #  * https://prometheus.io/docs/querying/rules/
 #  * https://prometheus.io/docs/practices/rules/
 #
-# NOTE: As of 2017-03, all rules are evalutated in parallel. Their evaluation
-# order is not guaranteed, and dependencies between rules are not respected.
-# Using recording rules on the right hand side of an expression can have
-# undefined behavior and may result in recording old data or other errors.
+# NOTE: As of 2017-11, the Prometheus v1.x series evaluates all rules in
+# parallel. So, rule evaluation order is not guaranteed, and dependencies
+# between rules are not respected. Using recording rules on the right hand
+# side of an expression can have undefined behavior and may result in recording
+# old data or other errors. This is also true for Alerts. This limitation is
+# fixed in the Prometheus v2.x series.
 #
-#    https://github.com/prometheus/prometheus/blob/master/rules/manager.go#L240
+# TODO: Prometheus v2.x rules are evaluated in order.
+#
+#    https://github.com/prometheus/prometheus/blob/v1.8.2/rules/manager.go#L254
 #
 # DO:
 #  * Do use raw prometheus expressions on the right hand side of a new rule.
+#  * "Recording rules should be of the general form level:metric:operations."
 #
 # DO NOT:
 #  * Do not use recording rules on the right hand side of a new rule.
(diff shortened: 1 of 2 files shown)


gfr10598:
Review status: 0 of 2 files reviewed at latest revision, 5 unresolved discussions.

---

*[config/federation/prometheus/rules.yml, line 39 at r1](https://reviewable.io:443/reviews/m-lab/prometheus-support/145#-L-Yj9i1U7Cru-Ti3rtn:-L-Yj9i1U7Cru-Ti3rto:b-u8eqt2) ([raw file](https://github.com/m-lab/prometheus-support/blob/2574fcfda5a8e2ad7973edf7cc58c7f996413177/config/federation/prometheus/rules.yml#L39)):*
> ```YAML
> 
> # Precalculate the sum of sidestream connections per experiment "last six bits".
> lsb:sidestream_connection_count:increase2m =
> ```

BTW: suggestion changing this to l6b if that works.

---

*[config/federation/prometheus/rules.yml, line 49 at r1](https://reviewable.io:443/reviews/m-lab/prometheus-support/145#-L-YkLn5D6pWRzPEB8KN:-L-YkLn5D6pWRzPEB8KO:b-lxwk2k) ([raw file](https://github.com/m-lab/prometheus-support/blob/2574fcfda5a8e2ad7973edf7cc58c7f996413177/config/federation/prometheus/rules.yml#L49)):*
> ```YAML
> machine:inotify_extension_create:rate2m =
>     # NOTE: using 'without' instead of 'by' preserves all other labels.
>     60.0 * sum without(ext) (rate(inotify_extension_create_total{ext=~".*_snaplog"}[2m]))
> ```

Leave out 60.0 *, or reflect it in the name - maybe rpm2m ?

---

*[config/federation/prometheus/rules.yml, line 53 at r1](https://reviewable.io:443/reviews/m-lab/prometheus-support/145#-L-Yk31drJNay62MVuek:-L-Yk31drJNay62MVuel:bf8ukqy) ([raw file](https://github.com/m-lab/prometheus-support/blob/2574fcfda5a8e2ad7973edf7cc58c7f996413177/config/federation/prometheus/rules.yml#L53)):*
> ```YAML
> # Per-switch "Out" (i.e. Download) bits per second.
> uplink:ifHCOutOctets:bits_per_second_rate2m =
>     8 * rate(ifHCOutOctets{ifAlias="uplink"}[2m])
> ```

Is the uplink alias preserved if we leave it out here?  Should we just collect this for all the links, or are you concerned about volume?
These rules basically double the timeseries volume, right?

---

*[config/federation/prometheus/rules.yml, line 55 at r1](https://reviewable.io:443/reviews/m-lab/prometheus-support/145#-L-YjR8FslMYsIw_74w7:-L-YjR8GHCRNYS5S4bU1:b-xotqrv) ([raw file](https://github.com/m-lab/prometheus-support/blob/2574fcfda5a8e2ad7973edf7cc58c7f996413177/config/federation/prometheus/rules.yml#L55)):*
> ```YAML
>     8 * rate(ifHCOutOctets{ifAlias="uplink"}[2m])
> 
> # Ratio of time spent performing I/O on the /vservers partition.
> ```

BTW: we should probably collect this for all partitions.  If they are on the same spindle, they will all impact disk performance.   So it would be a pity to overlook some IOs on a different partition.

---

*[config/federation/prometheus/rules.yml, line 57 at r1](https://reviewable.io:443/reviews/m-lab/prometheus-support/145#-L-Yjk2urMT8OwlU7CiP:-L-Yjk2urMT8OwlU7CiQ:b-h3qhug) ([raw file](https://github.com/m-lab/prometheus-support/blob/2574fcfda5a8e2ad7973edf7cc58c7f996413177/config/federation/prometheus/rules.yml#L57)):*
> ```YAML
> # Ratio of time spent performing I/O on the /vservers partition.
> vserver_partition:node_disk_io_time_ms:ratio_rate2m =
>     rate(node_disk_io_time_ms{device="dm-2"}[2m]) / 1000
> ```

Wondering if it is better to leave scaling by 1000 (or by 8 above) to the later queries instead of here.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/m-lab/prometheus-support/145)*

stephen-soltesz:
Your feedback is very helpful. Thank you. PTAL?

---

Review status: 0 of 2 files reviewed at latest revision, 5 unresolved discussions.

---

*[config/federation/prometheus/rules.yml, line 39 at r1](https://reviewable.io:443/reviews/m-lab/prometheus-support/145#-L-Yj9i1U7Cru-Ti3rtn:-L-YrIhb1R2QFBlTpbLR:b-j2znko) ([raw file](https://github.com/m-lab/prometheus-support/blob/2574fcfda5a8e2ad7973edf7cc58c7f996413177/config/federation/prometheus/rules.yml#L39)):*

BTW: suggestion changing this to l6b if that works.
</blockquote></details>

This is out of scope for this PR. Some dashboards already use this metric, and I would rather defer taking the time to make cascading changes to a time when we can reconsider how to map the "l6b" or "port range" to human readable names. That mapping must be solved on the node itself. And, these metrics currently are harder to use since the last six bits are not shared universally by all experiments.

---

*[config/federation/prometheus/rules.yml, line 49 at r1](https://reviewable.io:443/reviews/m-lab/prometheus-support/145#-L-YkLn5D6pWRzPEB8KN:-L-YlWoisE9eEMC5tLS_:bbalcvg) ([raw file](https://github.com/m-lab/prometheus-support/blob/2574fcfda5a8e2ad7973edf7cc58c7f996413177/config/federation/prometheus/rules.yml#L49)):*

Leave out 60.0 *, or reflect it in the name - maybe rpm2m ?
</blockquote></details>

Oh, good catch. I tried to do this for the uplink conversion to bits per second, but forgot here. "rpm2m" implies that rate is involved, so dropping the "rate" here in favor of just rpm2m seems clearer and more concise. :+1:

---

*[config/federation/prometheus/rules.yml, line 53 at r1](https://reviewable.io:443/reviews/m-lab/prometheus-support/145#-L-Yk31drJNay62MVuek:-L-YlrPKQnG5vRg4Revc:b-t3hkxu) ([raw file](https://github.com/m-lab/prometheus-support/blob/2574fcfda5a8e2ad7973edf7cc58c7f996413177/config/federation/prometheus/rules.yml#L53)):*

Is the uplink alias preserved if we leave it out here?  Should we just collect this for all the links, or are you concerned about volume?
These rules basically double the timeseries volume, right?
</blockquote></details>

Yes, the ifAlias would be preserved and this would double the timeseries. Though the uplink alias is the only meaningful name shared across all switches. I imagine that the per-server interfaces would go unused until we added meaningful names to them, which requires a switch config update. My resistance mainly comes from expecting that data going unused and it being difficult to work with if we tried. Right now it feels like a waste.

I'll ask Nathan about adding per-machine interface aliases. Added a TODO here and used a more generic level name that could be updated transparently with additional ifAliases.

---

*[config/federation/prometheus/rules.yml, line 55 at r1](https://reviewable.io:443/reviews/m-lab/prometheus-support/145#-L-YjR8FslMYsIw_74w7:-L-YsMtKziQSJiOYzCUc:blo84ph) ([raw file](https://github.com/m-lab/prometheus-support/blob/2574fcfda5a8e2ad7973edf7cc58c7f996413177/config/federation/prometheus/rules.yml#L55)):*

BTW: we should probably collect this for all partitions.  If they are on the same spindle, they will all impact disk performance.   So it would be a pity to overlook some IOs on a different partition.
</blockquote></details>

Yes. You're right. Good question. I thought this would be a much harder problem. These "dm-*" devices are LVM partitions over physical disks. And, while the more hardware we upgrade the closer we're getting to one physical disk with a predictable name, in the long tail and general case there could always be more than one disk. Apparently, at least 3 sites have nodes with 4 physical disks :-) And, I'm not sure what scheme we'll use on the new platform...

That said though -- The common case is and will probably remain a single physical disk. And, using `max` I think we can still find periods of extreme I/O irrespective of the device. This will tell us if any device is experiencing high sustained load. And, we can use the raw metrics to find out which device if needed.

There's a risk of some false positives from load that wouldn't affect test traffic, but we can evaluate that if it occurs.

How does this sound?

---

*[config/federation/prometheus/rules.yml, line 57 at r1](https://reviewable.io:443/reviews/m-lab/prometheus-support/145#-L-Yjk2urMT8OwlU7CiP:-L-Yn3bOgF6kWNxn5mbX:b-lcw3m8) ([raw file](https://github.com/m-lab/prometheus-support/blob/2574fcfda5a8e2ad7973edf7cc58c7f996413177/config/federation/prometheus/rules.yml#L57)):*

Wondering if it is better to leave scaling by 1000 (or by 8 above) to the later queries instead of here.
</blockquote></details>

Good question. My reasoning is that the compound queries we'll be writing that use these metrics will be complex enough without adding additional scaling operations, but that may not be true in general. In these specific cases, for ifHCOutOctets, I think "bits per second" is the most common unit for data rates. And, for `node_disk_io_time_ms` the unit resulting from `rate(node_disk_io_time_ms[2m])` is "ms / s" by default. The "/1000" normalizes to "s/s", which seems preferable, since this has a direct interpretation as a percentage.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/m-lab/prometheus-support/145#-:-L-Z10jBiL6Pxi3C4-ps:b20cgbg)*

gfr10598:
:lgtm:

---

Review status: 0 of 2 files reviewed at latest revision, 4 unresolved discussions.

---

*[config/federation/prometheus/rules.yml, line 55 at r1](https://reviewable.io:443/reviews/m-lab/prometheus-support/145#-L-YjR8FslMYsIw_74w7:-L-ZuxHfW2W4dpmUZVAP:b-jnwmmd) ([raw file](https://github.com/m-lab/prometheus-support/blob/2574fcfda5a8e2ad7973edf7cc58c7f996413177/config/federation/prometheus/rules.yml#L55)):*

Yes. You're right. Good question. I thought this would be a much harder problem. These "dm-*" devices are LVM partitions over physical disks. And, while the more hardware we upgrade the closer we're getting to one physical disk with a predictable name, in the long tail and general case there could always be more than one disk. Apparently, at least 3 sites have nodes with 4 physical disks :-) And, I'm not sure what scheme we'll use on the new platform...

That said though -- The common case is and will probably remain a single physical disk. And, using `max` I think we can still find periods of extreme I/O irrespective of the device. This will tell us if any device is experiencing high sustained load. And, we can use the raw metrics to find out which device if needed.

There's a risk of some false positives from load that wouldn't affect test traffic, but we can evaluate that if it occurs.

How does this sound?
</blockquote></details>

I'm most concerned if there are multiple partitions on one spindle, and we are only seeing a subset of the IO on that spindle.  We might then underestimate the utilization.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/m-lab/prometheus-support/145#-:-L-ZvMMx6UEGWcYdKF_Z:bnfp4nl)*

stephen-soltesz:
Review status: 0 of 2 files reviewed at latest revision, 4 unresolved discussions, all commit checks successful.

---

*[config/federation/prometheus/rules.yml, line 55 at r1](https://reviewable.io:443/reviews/m-lab/prometheus-support/145#-L-YjR8FslMYsIw_74w7:-L-bPKHPccisio6qCcXM:bo7gfn4) ([raw file](https://github.com/m-lab/prometheus-support/blob/2574fcfda5a8e2ad7973edf7cc58c7f996413177/config/federation/prometheus/rules.yml#L55)):*

I'm most concerned if there are multiple partitions on one spindle, and we are only seeing a subset of the IO on that spindle.  We might then underestimate the utilization.
</blockquote></details>

Right. I believe this approach addresses that concern at the minor risk of over estimating the utilization.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/m-lab/prometheus-support/145)*

PR #148, opened 2017-12-08
Enable more reliable recording rules using irate

stephen-soltesz:
The original recording rules resulted in a random set of about 20-30 missing switch uplink time series.

After describing my investigation here https://groups.google.com/forum/#!topic/prometheus-users/LVidx6lwwAo I got some very helpful feedback and came to a better solution.

In particular, this change uses `irate` instead of `rate` and a larger range window to guarantee that two points are always available for the `irate` calculation.

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/prometheus-support/148)


gfr10598:
:lgtm:
Glad you figured this out.

---







*Comments from [Reviewable](https://reviewable.io:443/reviews/m-lab/prometheus-support/148#-:-L-t2d7iS1ckdMnazPoa:bgignvm)*

PR #155, opened 2017-12-18
Increases timeout for tcp_v4_[tls_]online modules from 5s to 10s

nkinkade:
Intuition would dictate that 5s is more than enough to establish a TCP connection to a remote server, but for some reason a lgood number of the `ndt_ssl` and `ndt_raw` checks are timing out at 5s. In order to find out the depth of this problem, this PR increases the timeout to 10s. If things are still timing out at 10s, then something fundamental must be wrong which will warrant a more in-depth investigation. But currently, we don't know whether the timed-out check would have succeeded a few seconds later, or whether the problem is great enough that it would even timeout at 10s. After this PR we'll be able to assess the situation better.

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/prometheus-support/155)

### DIFF
--- a/config/federation/blackbox/config.yml
+++ b/config/federation/blackbox/config.yml
@@ -29,14 +29,14 @@ modules:
   # target=<hostname:port>
   tcp_v4_online:
     prober: tcp
-    timeout: 5s
+    timeout: 10s
     tcp:
       protocol: "tcp4"
 
   # target=<hostname:port>
   ssh_v4_online:
     prober: tcp
-    timeout: 10s
+    timeout: 5s
     tcp:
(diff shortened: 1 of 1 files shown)


stephen-soltesz:
I suggest making this 9s so that we can continue to distinguish between timeouts due to bbe vs prometheus (which is also 10s).

:lgtm:

---

Reviewed 1 of 1 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/m-lab/prometheus-support/155#-:-L0fdctzKCLTmIyjs1bA:b-7tat39)*

PR #159, opened 2017-12-21
Increase resource alloc for blackbox exporter

stephen-soltesz:
This change addresses a performance regression in the blackbox exporter.

We believe that the issue is due to the Go GC as originally reported in https://github.com/prometheus/blackbox_exporter/issues/270

It may be that one day the Go compiler will fix the inefficiency but until then, we can simply provide more resources.

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/prometheus-support/159)

### DIFF
--- a/k8s/prometheus-federation/deployments/blackbox.yml
+++ b/k8s/prometheus-federation/deployments/blackbox.yml
@@ -41,12 +41,15 @@ spec:
         ports:
           - containerPort: 9115
         resources:
+          # TODO: reduce resource allocation once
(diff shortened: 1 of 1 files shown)


nkinkade:
:lgtm:

---

Reviewed 1 of 1 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/m-lab/prometheus-support/159#-:-L0uqPfFvgdYSfrt6eDu:bnfp4nl)*

PR #176, opened 2018-01-23
Enable better memory management by Prometheus

stephen-soltesz:
This change adds new flags to the prometheus-federation k8s prometheus deployment. In particular, the `target-heap-size` is set to 75% (9GB) of the currently configured RAM for all three projects (12GB).

This configuration was deployed to mlab-oti as a hotfix today to address increasingly slow behavior that started with the production tag Monday that included the new ParserDailyVolumeTooLow query.

The other flag values are 10x the default values and are the same as those used by the scraper-cluster configuration.

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/prometheus-support/176)


gfr10598:
:lgtm:

---

Review status: 0 of 2 files reviewed at latest revision, 1 unresolved discussion.

---

*[.travis.yml, line 96 at r1](https://reviewable.io:443/reviews/m-lab/prometheus-support/176#-L3_7VllaMB4JPdkJZwK:-L3_7VllaMB4JPdkJZwL:b-qmr8n0) ([raw file](https://github.com/m-lab/prometheus-support/blob/4eef7977c9983d892d310fa94b9519c08ebf48e0/.travis.yml#L96)):*
> ```YAML
>     all_branches: true
>     # Note: condition allows pushes to a hotfix branch or tags that match a prefix "production"
>     condition: ( $TRAVIS_EVENT_TYPE == push && $TRAVIS_BRANCH == hotfix-* ) || $TRAVIS_TAG == production*
> ```

Do you still want the hotfix filter, or was that just to perform the hotfix today?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/m-lab/prometheus-support/176#-:-L3_7e1g_TnTUPkr9QX6:bnfp4nl)*

stephen-soltesz:
Review status: 0 of 2 files reviewed at latest revision, 1 unresolved discussion, some commit checks pending.

---

*[.travis.yml, line 96 at r1](https://reviewable.io:443/reviews/m-lab/prometheus-support/176#-L3_7VllaMB4JPdkJZwK:-L3_B3KGb-OoNtbHTEJh:b-j5iiuf) ([raw file](https://github.com/m-lab/prometheus-support/blob/4eef7977c9983d892d310fa94b9519c08ebf48e0/.travis.yml#L96)):*

Do you still want the hotfix filter, or was that just to perform the hotfix today?
</blockquote></details>

My thinking here is that we want to support some mechanism for deploying hot fixes in case of emergency. For example, using a branch rather than another special tag is that if we're actively trying to diagnose a thing, we may need to deploy multiple changes in succession. And, a branch is better suited to that. I also don't want someone to have to reinvent a fast deployment mechanism and risk making things worse. Although that is a risk with any direct push-to-prod mechanism. I'm open to more discussion.

I've added an additional comment warning that hotfix branches should only  be used in emergencies.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/m-lab/prometheus-support/176)*

PR #189, opened 2018-02-06
Add explicit version of gcp-service-discovery docker image

stephen-soltesz:
This change adds an explicit version to the gcp-service-discovery docker image.

After fixing a bug earlier today, the "latest" tag in dockerhub is replaced with the new "latest" version. This was fine, but it means if the staging or production instances are restarted they will get the updated version without notice.

By adding an explicit version we can prevent unintended side effects on production environments.

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/prometheus-support/189)

### DIFF
--- a/k8s/prometheus-federation/deployments/prometheus.yml
+++ b/k8s/prometheus-federation/deployments/prometheus.yml
@@ -133,7 +133,7 @@ spec:
         - mountPath: /prometheus
           name: prometheus-storage
 
-      - image: measurementlab/gcp-service-discovery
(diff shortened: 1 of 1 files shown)


nkinkade:
:lgtm:

---

Review status: 0 of 1 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/m-lab/prometheus-support/189#-:-L4cPvM1FR-_AdofmWs2:bnfp4nl)*

PR #198, opened 2018-03-14
Configures IPv6 modules for the blackbox_exporter

nkinkade:
This PR also configures gcp-service-discovery to fetch IPv6 targets for mobiperf and neubot. These targets will be generated by the changes in m-lab/operator#205.

Our current Nagios configuration checks both IPv4 and IPv6 for neubot and mobiperf. In order to keep parity in our migration from Nagios to Prometheus, we should do IPv6 checks for these services in Prometheus too. In order to achieve this we also need blackbox_exporter modules for IPv4 checks. This PR implements these.

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/prometheus-support/198)

### DIFF
--- a/config/federation/blackbox/config.yml
+++ b/config/federation/blackbox/config.yml
@@ -40,12 +40,21 @@ modules:
     tcp:
       preferred_ip_protocol: "ip6"
 
-  # target=<hostname:port>
+  # IPv4 target=<hostname:port>
   ssh_v4_online:
     prober: tcp
     timeout: 9s
     tcp:
       preferred_ip_protocol: "ip4"
+      query_response:
+        - expect: "SSH-2.0-OpenSSH_.+"
+
+  # IPv6 target=<hostname:port>
+  ssh_v6_online:
+    prober: tcp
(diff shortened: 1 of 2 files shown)


nkinkade:
I've added another commit which adds a new IPv6 module blackbox_exporter for the ssh806 service. This was at the request of @mattmathis so that we'd have a way of checking IPv6 connectivity at a site that was independent of experiments.

stephen-soltesz:
:lgtm: -- this looks okay, but I'm guessing ipv6 checks won't work from k8s / GCE vms?

---

Review status: 0 of 2 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/m-lab/prometheus-support/198#-:-L7eMCNQDm6_-gcccwk-:b-t001q3)*

nkinkade:
I forgot that GCP doesn't really support IPv6  (except on Load Balancers). In light of that, this PR doesn't make any sense. :(

PR #216, opened 2018-04-06
Add basic auth credentials to alertmanager URL

jheretic:
This prepends the basic auth credentials for prometheus to the alertmanager URL, so that alerts sent to Slack can be clicked on directly and be authenticated without additional user interaction. Accordingly, the way that the basic authorization string is generated is changed. The string is now generated by Travis, using plaintext username and password added as separate environment variables in the Travis environment and the `htpasswd` utility that has been added to Travis. The alertmanager pod must be restarted before the modified URLs are used.

In addition to the above changes, the `proxy-connect-timeout` for the nginx ingress is increased to 300 seconds.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -160,3 +160,5 @@ deploy:
 before_install:
 - travis/decrypt.sh "$encrypted_1c8dd3704323_key" "$encrypted_1c8dd3704323_iv"
   keys/service-accounts.tar.enc /tmp/service-accounts.tar /tmp
+- sudo apt-get -qq update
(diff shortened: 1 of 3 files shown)


stephen-soltesz:
:lgtm: - Thanks!

---

Reviewed 3 of 3 files at r1.
Review status: all files reviewed at latest revision, all discussions resolved, some commit checks failed.

---



*Comments from [Reviewable](https://beta.reviewable.io/reviews/m-lab/prometheus-support/216#-:-L9RvQ_pEEGDoKyuZqhH:ba0pql2)*

PR #222, opened 2018-04-30
Adds a scrape job for the node_exporter running on eb.measurementlab.net

nkinkade:
node_exporter was installed on EB via the system package manager. It's a pretty old version (0.13), but sufficient for monitoring basic memory, load and disk usage.

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/prometheus-support/222)

PR #223, opened 2018-05-03
Alerting for BBE when the BBE service itself is down, not just probes.

nkinkade:
In PR #222 I attempted to add alerting for all of our Prometheus blackbox_exporter instances. But what I inadvertently did was to add alerting for individual BBE probes, not the services themselves. This PR should resolve that error.

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/prometheus-support/223)

### DIFF
--- a/apply-global-prometheus.sh
+++ b/apply-global-prometheus.sh
@@ -25,6 +25,18 @@ CLUSTER=${CLUSTER:?Please provide cluster name: $USAGE}
 
 export GRAFANA_DOMAIN=grafana.${PROJECT}.measurementlab.net
 
+# GCP doesn't support IPv6, so we have a Linode VM running three instances of
+# the blackbox_exporter, on three separate ports... one port/instance for each
+# project. These variables map projects to ports.
+BBE_IPV6_PORT_mlab_oti="9115"
+BBE_IPV6_PORT_mlab_staging="8115"
+BBE_IPV6_PORT_mlab_sandbox="7115"
+
+# Construct the per-project blackbox_exporter port using the passed $PROJECT
+# argument.
+bbe_port=BBE_IPV6_PORT_${PROJECT/-/_}
+
+
 # Config maps and Secrets
 
 ## Blackbox exporter.
@@ -36,6 +48,7 @@ kubectl create configmap blackbox-config \
 
 # Evaluate the configuration template.
 sed -e 's|{{PROJECT}}|'${PROJECT}'|g' \
+    -e 's|{{BBE_IPV6_PORT}}|'${!bbe_port}'|g' \
     config/federation/prometheus/prometheus.yml.template > \
     config/federation/prometheus/prometheus.yml
 
--- a/config/federation/prometheus/alerts.yml
+++ b/config/federation/prometheus/alerts.yml
@@ -315,7 +315,8 @@ ALERT ScriptExporterMissingMetrics
 # Prometheus is unable to get data from the blackbox_exporter service for IPv4
 # probes.
 ALERT BlackboxExporterIpv4DownOrMissing
-  IF up{job="blackbox-targets"} == 0 OR absent(up{job="blackbox-targets"})
+  IF up{deployment="blackbox-server"} == 0
+        OR absent(up{deployment="blackbox-server"})
   FOR 10m
   LABELS {
     severity = "ticket"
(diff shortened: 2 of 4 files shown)


stephen-soltesz:
Reviewed 2 of 3 files at r1.
Review status: 2 of 3 files reviewed at latest revision, all discussions resolved.

---

*[config/federation/prometheus/alerts.yml, line 332 at r1](https://reviewable.io/reviews/m-lab/prometheus-support/223#-LBblk3hBUSwySGl3R1J:-LBbeQYO98RqdXFGxEzV:bz7cvmk) ([raw file](https://github.com/m-lab/prometheus-support/blob/8b4b1e21111808578bba7c0a79688e69068166a4/config/federation/prometheus/alerts.yml#L332)):*
> ```YAML
> # probes. The service is down, or the metric is missing.
> ALERT BlackboxExporterIpv6DownOrMissing
>   IF up{job="blackbox-exporter-ipv6"} == 0
> ```

Hmm, I'm not sure what the right solution to the following problem is.

It's weird to me that BlackboxExporterIpv4DownOrMissing uses `up{deployment=...}` and BlackboxExporterIpv6DownOrMissing uses `up{job=...}`.

Ideally, there would be some consistent pattern to "the server itself" and "individual targets". Right now this varies depending on whether there's an explicit job defined or whether the pod is auto-discovered by prometheus from k8s.

I would really like these to be consistent.

I can imagine that we disable auto-discovery of the pods using `prometheus.io/scrape: 'false'` in the blackbox xporter k8s config and explicitly add a new job definition for it with the name "blackbox-exporter-ipv4". Then the BlackboxExporterIpv4DownOrMissing alert would use the same pattern as the BlackboxExporterIpv6DownOrMissing alert.

What do you think about this?

---


*Comments from [Reviewable](https://reviewable.io/reviews/m-lab/prometheus-support/223)*

nkinkade:
@stephen-soltesz: I think I've addressed your concern. PTAL?

---

Review status: 1 of 4 files reviewed at latest revision, 1 unresolved discussion.

---

*[config/federation/prometheus/alerts.yml, line 332 at r1](https://reviewable.io/reviews/m-lab/prometheus-support/223#-LBblk3hBUSwySGl3R1J:-LBbxjGYEjOpmnwovmeU:b9gjx8z) ([raw file](https://github.com/m-lab/prometheus-support/blob/8b4b1e21111808578bba7c0a79688e69068166a4/config/federation/prometheus/alerts.yml#L332)):*

Hmm, I'm not sure what the right solution to the following problem is.

It's weird to me that BlackboxExporterIpv4DownOrMissing uses `up{deployment=...}` and BlackboxExporterIpv6DownOrMissing uses `up{job=...}`.

Ideally, there would be some consistent pattern to "the server itself" and "individual targets". Right now this varies depending on whether there's an explicit job defined or whether the pod is auto-discovered by prometheus from k8s.

I would really like these to be consistent.

I can imagine that we disable auto-discovery of the pods using `prometheus.io/scrape: 'false'` in the blackbox xporter k8s config and explicitly add a new job definition for it with the name "blackbox-exporter-ipv4". Then the BlackboxExporterIpv4DownOrMissing alert would use the same pattern as the BlackboxExporterIpv6DownOrMissing alert.

What do you think about this?
</blockquote></details>

I think this sounds fine, and I've made a couple of new commits that I think addresses this concern.

---


*Comments from [Reviewable](https://reviewable.io/reviews/m-lab/prometheus-support/223#-:-LBbxqTgA6q7zqlI2c2Z:b-2zhuud)*

stephen-soltesz:
:lgtm: -- Thank you for making the additional changes!

---

Review status: 1 of 4 files reviewed at latest revision, 1 unresolved discussion.

---

*[config/federation/prometheus/prometheus.yml.template, line 344 at r2](https://reviewable.io/reviews/m-lab/prometheus-support/223#-LBc6eTA3cSjlcD8bXS7:-LBc6eTA3cSjlcD8bXS8:b-hd3e6) ([raw file](https://github.com/m-lab/prometheus-support/blob/2e8f8abef9ab4dd25720181284c460236d10e288/config/federation/prometheus/prometheus.yml.template#L344)):*
> ```
>   #
>   # Scrape the blackbox_exporter instance running in the cluster for service
>   # metrics.
> ```

Please add an additional note that "We do not use k8s pod discovery so that we can control the job name."

---

*[k8s/prometheus-federation/deployments/blackbox.yml, line 25 at r2](https://reviewable.io/reviews/m-lab/prometheus-support/223#-LBc6GhX41odmr6AhTTy:-LBc6GhX41odmr6AhTTz:b-66au4r) ([raw file](https://github.com/m-lab/prometheus-support/blob/2e8f8abef9ab4dd25720181284c460236d10e288/k8s/prometheus-federation/deployments/blackbox.yml#L25)):*
> ```YAML
>         run: blackbox-server
>       annotations:
>         # Tell prometheus service discovery to scrape the blackbox container.
> ```

Please update the comment. Please add a short (one sentence) explaining that the pod is scraped via an explicit job in prometheus.yml config.

---


*Comments from [Reviewable](https://reviewable.io/reviews/m-lab/prometheus-support/223#-:-LBc6rPG-yf7Fmp_0pev:bp233wy)*

PR #224, opened 2018-05-09
Adds an alert for the collectd-mlab service.

nkinkade:
This change is [Reviewable](https://reviewable.io/reviews/m-lab/prometheus-support/224)

### DIFF
--- a/config/federation/prometheus/alerts.yml
+++ b/config/federation/prometheus/alerts.yml
@@ -43,9 +43,10 @@ ALERT ClusterDown
 
 # MachineSLO
 #
-# SidestreamIsNotRunning: an M-Lab server is online, but the sidestream exporter
-# is not. Since sidestream is a core service, this must be fixed.
-ALERT SidestreamIsNotRunning
+# CoreServices_SidestreamIsNotRunning: an M-Lab server is online, but the
+# sidestream exporter is not. Since sidestream is a core service, this must be
+# fixed.
+ALERT CoreServices_SidestreamIsNotRunning
   IF sum_over_time(up{service="sidestream"}[10m]) == 0
         AND ON(machine)
      sum_over_time(probe_success{service="ssh806"}[20m]) / 20 >= 0.90
@@ -441,9 +442,10 @@ ALERT LameDuckMetricMissingForNode
   }
 
 # The collectd-mlab service has a problem.
-ALERT CollectdMlabDownOrMissing
+ALERT Coreservices_CollectdMlabDownOrMissing
   IF collectd_mlab_success{} == 0
-        OR absent(collectd_mlab_success{})
+        OR up{service="nodeexporter"} == 1
+            UNLESS ON(machine) collectd_mlab_success{}
   FOR 10m
   LABELS {
     severity = "ticket"
--- a/config/federation/prometheus/alerts.yml
+++ b/config/federation/prometheus/alerts.yml
@@ -441,17 +441,28 @@ ALERT LameDuckMetricMissingForNode
     hints = "Check /var/spool/node_exporter/ on the node to see if the file lame_duck.prom is missing. If it is, use the mlabops Ansible lame-duck.yaml playbook to restore it."
   }
 
-# The collectd-mlab service has a problem.
-ALERT Coreservices_CollectdMlabDownOrMissing
+# A collectd-mlab service has a problem and is down.
+ALERT Coreservices_CollectdMlabDown
   IF collectd_mlab_success{} == 0
-        OR up{service="nodeexporter"} == 1
-            UNLESS ON(machine) collectd_mlab_success{}
   FOR 10m
   LABELS {
     severity = "ticket"
   }
   ANNOTATIONS {
-    summary = "The collectd-mlab service is either down or the metric missing.",
+    summary = "A collectd-mlab service is down.",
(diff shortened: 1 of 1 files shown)


stephen-soltesz:
Review status: 0 of 1 files reviewed at latest revision, all discussions resolved.

---

*[config/federation/prometheus/alerts.yml, line 444 at r1](https://reviewable.io/reviews/m-lab/prometheus-support/224#-LC5yyEf2mjIFGYzkk5S:-LC5yyEf2mjIFGYzkk5T:b-awaaze) ([raw file](https://github.com/m-lab/prometheus-support/blob/c2b3925b38c791b83b9c74d7a14fa5375cba51be/config/federation/prometheus/alerts.yml#L444)):*
> ```YAML
> 
> # The collectd-mlab service has a problem.
> ALERT CollectdMlabDownOrMissing
> ```

We have enough alerts now that it may be helpful to being prefixing them with some context, like:
```
CoreServices_CollectdMlabDownOrMissing
```

It could also make sense to name the SidestreamIsNotRunning alert in the same way (would you be willing to make that change in this PR?). And, please open an issue to create a new check for paris traceroute. That would complete the set of "core services" monitoring. Ages ago, I made the assertion that we should not serve traffic on servers without the core services running. The next natural step would be to integrate the core-service checks into the mlab-ns queries.

---

*[config/federation/prometheus/alerts.yml, line 446 at r1](https://reviewable.io/reviews/m-lab/prometheus-support/224#-LC5zFyZAxTrfsf1je_t:-LC5zFyZAxTrfsf1je_u:b-27v2kn) ([raw file](https://github.com/m-lab/prometheus-support/blob/c2b3925b38c791b83b9c74d7a14fa5375cba51be/config/federation/prometheus/alerts.yml#L446)):*
> ```YAML
> ALERT CollectdMlabDownOrMissing
>   IF collectd_mlab_success{} == 0
>         OR absent(collectd_mlab_success{})
> ```

The absent check will fire only if all machines are not reporting the value, is that right?

Is there a way to enforce the absence check (without explicitly enumerating all hosts)?

Could we construct an `ON(machine)` query against the node exporters (which are all pre-configured) that returns only those missing the `collectd_mlab_success` metric? -- I haven't thought about this very hard so the answer may be  "no". But if so, this would better fulfill the promise of the alert name.

---


*Comments from [Reviewable](https://reviewable.io/reviews/m-lab/prometheus-support/224)*

stephen-soltesz:
:lgtm: -- please split the "down" and "missing" checks.

---

Review status: 0 of 1 files reviewed at latest revision, all discussions resolved.

---

*[config/federation/prometheus/alerts.yml, line 445 at r2](https://reviewable.io/reviews/m-lab/prometheus-support/224#-LCVmH_SDk_AnBx1MWCo:-LCVmH_SDk_AnBx1MWCp:b-93c19l) ([raw file](https://github.com/m-lab/prometheus-support/blob/3ba579c73fc95611870c8e9dd8f0fb15f784f6d1/config/federation/prometheus/alerts.yml#L445)):*
> ```YAML
> 
> # The collectd-mlab service has a problem.
> ALERT Coreservices_CollectdMlabDownOrMissing
> ```

Something has worried me for a while about combining "down" & "missing" alerts. For singleton services I've ignored my feeling. For alerts on sets of services, I am uncomfortable enough that I think we should split this into two alerts.

`Coreservices_CollectdMlabDown`
`Coreservices_CollectdMlabMissing`

These are distinct situations and likely have different resolutions.

I'm also not confident about the meaning of `A == 0 or B unless A` . I think each alert is simpler and easier to reason about when each is independent.

Please split these.

---


*Comments from [Reviewable](https://reviewable.io/reviews/m-lab/prometheus-support/224#-:-LCVoToI4QgzH5K5J1ah:b-todj1s)*

nkinkade:
They are now split.

---

Review status: 0 of 1 files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---

*[config/federation/prometheus/alerts.yml, line 445 at r2](https://reviewable.io/reviews/m-lab/prometheus-support/224#-LCVmH_SDk_AnBx1MWCo:-LCVsujLA26oCygKVRma:b-896fix) ([raw file](https://github.com/m-lab/prometheus-support/blob/3ba579c73fc95611870c8e9dd8f0fb15f784f6d1/config/federation/prometheus/alerts.yml#L445)):*

Something has worried me for a while about combining "down" & "missing" alerts. For singleton services I've ignored my feeling. For alerts on sets of services, I am uncomfortable enough that I think we should split this into two alerts.

`Coreservices_CollectdMlabDown`
`Coreservices_CollectdMlabMissing`

These are distinct situations and likely have different resolutions.

I'm also not confident about the meaning of `A == 0 or B unless A` . I think each alert is simpler and easier to reason about when each is independent.

Please split these.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io/reviews/m-lab/prometheus-support/224#-:-LCVsxJ--M52wQfbQYLC:bebw7od)*

nkinkade:
Oh, and I added this issue, too:

https://github.com/m-lab/prometheus-support/issues/225

---

---



*Comments from [Reviewable](https://reviewable.io/reviews/m-lab/prometheus-support/224#-:-LCVuijfBHZ0gDAs3evI:bxmady2)*

PR #239, opened 2018-06-07
Adds two new alerts: vdlimit_* metrics and SNMP scraping

nkinkade:
* mlab-ns relies on three specific metrics from the node_exporter instance running on each node: `vdlimit_used`, `vdlimit_total`, `lame_duck_experiment`. There is already an alert for `lame_duck_experiment`. This PR adds an alert for both of the vdlimit_* metrics.

* If SNMP scraping of a switch is failing, we will now get an alert. Currently, we only get an alert if snmp_exporter is down or if an entire metric is missing from all switches.

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/prometheus-support/239)

### DIFF
--- a/config/federation/prometheus/alerts.yml
+++ b/config/federation/prometheus/alerts.yml
@@ -310,7 +310,8 @@ ALERT SnmpScrapingDownAtSite
   }
   ANNOTATIONS {
     summary = "Prometheus is unable to scrape SNMP metrics from a switch.",
-    hints = "Maybe the switch is down? Is the snmp_exporter using the right community string? Look in switch-details.json in the m-lab/switch-config repo. Is the IP of the snmp_exporter VM in GCE whitelisted on the switch?"
+    hints = "Maybe the switch is down? Is the snmp_exporter using the right community string? Look in switch-details.json in the m-lab/switch-config repo. Is the IP of the snmp_exporter VM in GCE whitelisted on the switch?",
+    dashboard = "https://grafana.mlab-oti.measurementlab.net/dashboard/file/Ops_SwitchOverview.json?orgId=1&var-site_name={{ $labels.site }}"
   }
 
 # Prometheus is unable to get data from the script_exporter service.
@@ -492,7 +493,8 @@ ALERT VdlimitMetricsMissingForNode
   }
   ANNOTATIONS {
     summary = "Some vdlimit_* metrics are missing.",
-    hints = "Check /var/spool/node_exporter/ on the node to see if the file vdlimit.prom is missing. The file is created by /etc/cron.d/prom_vdlimit_metrics.cron."
+    hints = "Check /var/spool/node_exporter/ on the node to see if the file vdlimit.prom is missing. The file is created by /etc/cron.d/prom_vdlimit_metrics.cron.",
+    dashboard = "https://grafana.mlab-oti.measurementlab.net/dashboard/file/Ops_PlatformOverview.json?orgId=1"
   }
 
 # A collectd-mlab service has a problem and is down.
--- a/config/federation/prometheus/alerts.yml
+++ b/config/federation/prometheus/alerts.yml
@@ -309,7 +309,7 @@ ALERT SnmpScrapingDownAtSite
     repo = "ops-tracker"
   }
   ANNOTATIONS {
-    summary = "Prometheus is unable to scrape SNMP metrics from a switch."
+    summary = "Prometheus is unable to scrape SNMP metrics from a switch.",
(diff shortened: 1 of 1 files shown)


stephen-soltesz:
Review status: 0 of 1 files reviewed at latest revision, all discussions resolved, some commit checks broke.

---

*[config/federation/prometheus/alerts.yml, line 313 at r1](https://reviewable.io/reviews/m-lab/prometheus-support/239#-LERB8uqE11qI5gGQ-NR:-LERB8uqE11qI5gGQ-NS:bgjlvy2) ([raw file](https://github.com/m-lab/prometheus-support/blob/4a709e7e1b823346830a097ac92ef1479dc4b031/config/federation/prometheus/alerts.yml#L313)):*
> ```YAML
>   ANNOTATIONS {
>     summary = "Prometheus is unable to scrape SNMP metrics from a switch."
>     hints = "Maybe the switch is down? Is the snmp_exporter using the right community string? Look in switch-details.json in the m-lab/switch-config repo. Is the IP of the snmp_exporter VM in GCE whitelisted on the switch?"
> ```

Suggest adding a link to the switch overview dashboard: https://grafana.mlab-sandbox.measurementlab.net/dashboard/file/Ops_SwitchOverview.json?orgId=1

This could help investigate when exactly the site went offline.

I've not tested this before but it should be possible to use a template value like:

`<above url>&var-site_name={{ $labels.site }}`

To construct a URL that is specific to the site. -- this may not work when there are multiple sites failing at the same time. We'd have to debug it. But, that can be for a future refinement. For now, just the link above is enough.

---

*[config/federation/prometheus/alerts.yml, line 495 at r1](https://reviewable.io/reviews/m-lab/prometheus-support/239#-LERCTtj6AgL2qE9w6kj:-LERCTtj6AgL2qE9w6kk:b-hw91uj) ([raw file](https://github.com/m-lab/prometheus-support/blob/4a709e7e1b823346830a097ac92ef1479dc4b031/config/federation/prometheus/alerts.yml#L495)):*
> ```YAML
>   ANNOTATIONS {
>     summary = "Some vdlimit_* metrics are missing.",
>     hints = "Check /var/spool/node_exporter/ on the node to see if the file vdlimit.prom is missing. The file is created by /etc/cron.d/prom_vdlimit_metrics.cron."
> ```

Maybe a link to the ops platform overview dashboard?

I think it will also be possible to reference these hint urls from the Slack text notification message, so that the link could be immediately presented at the same time an alert fires.

---


*Comments from [Reviewable](https://reviewable.io/reviews/m-lab/prometheus-support/239)*

stephen-soltesz:
:lgtm: -- with suggested additions.

---

Review status: 0 of 1 files reviewed at latest revision, 2 unresolved discussions, some commit checks broke.

---



*Comments from [Reviewable](https://reviewable.io/reviews/m-lab/prometheus-support/239#-:-LERClPi0pykP9AdFDYg:b-2pwojn)*

nkinkade:
Review status: 0 of 1 files reviewed at latest revision, 2 unresolved discussions, all commit checks successful.

---

*[config/federation/prometheus/alerts.yml, line 313 at r1](https://reviewable.io/reviews/m-lab/prometheus-support/239#-LERB8uqE11qI5gGQ-NR:-LERJTUn71iq_ShpKUcQ:b-896fix) ([raw file](https://github.com/m-lab/prometheus-support/blob/4a709e7e1b823346830a097ac92ef1479dc4b031/config/federation/prometheus/alerts.yml#L313)):*

Suggest adding a link to the switch overview dashboard: https://grafana.mlab-sandbox.measurementlab.net/dashboard/file/Ops_SwitchOverview.json?orgId=1

This could help investigate when exactly the site went offline.

I've not tested this before but it should be possible to use a template value like:

`<above url>&var-site_name={{ $labels.site }}`

To construct a URL that is specific to the site. -- this may not work when there are multiple sites failing at the same time. We'd have to debug it. But, that can be for a future refinement. For now, just the link above is enough.
</blockquote></details>

Done.

---

*[config/federation/prometheus/alerts.yml, line 495 at r1](https://reviewable.io/reviews/m-lab/prometheus-support/239#-LERCTtj6AgL2qE9w6kj:-LERJTwK2WXPBd5XZPT5:b-896fix) ([raw file](https://github.com/m-lab/prometheus-support/blob/4a709e7e1b823346830a097ac92ef1479dc4b031/config/federation/prometheus/alerts.yml#L495)):*

Maybe a link to the ops platform overview dashboard?

I think it will also be possible to reference these hint urls from the Slack text notification message, so that the link could be immediately presented at the same time an alert fires.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io/reviews/m-lab/prometheus-support/239)*

PR #248, opened 2018-06-18
Run blackbox exporter redundantly

stephen-soltesz:
This change runs the blackbox exporter redundantly and enables pod- service discovery for the individual instances. It also preserves monitoring of the `job_name: 'blackbox-exporter-ipv4'` target so that our alerts enforce that there is at least one available, but all metrics from the service target are discarded (since they would not be meaningful anyway).

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/prometheus-support/248)

### DIFF
--- a/k8s/prometheus-federation/deployments/blackbox.yml
+++ b/k8s/prometheus-federation/deployments/blackbox.yml
@@ -25,6 +25,19 @@ spec:
         # Tell prometheus service discovery to scrape the blackbox container.
         prometheus.io/scrape: 'true'
     spec:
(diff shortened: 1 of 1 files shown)


nkinkade:
:lgtm:

---

Reviewed 2 of 2 files at r1.
Review status: :shipit: complete! 1 of 1 LGTMs obtained

---



*Comments from [Reviewable](https://reviewable.io/reviews/m-lab/prometheus-support/248#-:-LFNitFx35bpxNv2ExPG:bnfp4nl)*

Issue #263, opened 2018-06-26
Update alert annotation URLs for Grafana 5.x URLs


stephen-soltesz:
Grafana 5.x uses a new URL format for json and native dashboards.

Any links used in the promethues/alerts.yml annotations will need to be updated to the new form.

PR #265, opened 2018-06-27
Fix dashboard URLs in alerts.yaml + updates JSON dashboards to use 5.x formatting

nkinkade:
Currently, all of the Grafana dashboard URLs in alerts.yaml are using an old v4.x format, but we are now running v5.x which has a different URL format for resources. This PR fixes that. It also updates several other URLs which were broken or non-standard in some way e.g., using http instead of https, using `status.<url>` URLs instead of `prometheus.<url>`.

Additionally, with v5.x, Grafana now formats the JSON for a dashboard significantly different. This PR updates the format. This was done by simply opening the dashboard JSON in Grafana and pasting it into the JSON file in clone of the repo. The diffs are large, but the dashboards should be functionally and visually the same. The only dashboard which I altered was Ops_PodOverview, and this was to take advantage of the new layout flexibility of Grafana 5.x, and to add two new singlestat panels for switch-snmp and switch-ping for a pod.

Also of note in this PR, is that the `[paths] provisioning = /etc/provisioning` configuration was removed from `grafana.ini` and the configuration moved to an environment variable in the k8s config for the Grafana container. This is because the Docker images the Grafana project publishes [set various environment variables which override anything in the config file](https://github.com/grafana/grafana-docker/blob/master/Dockerfile#L13). The only way to override their overrides is with an environment variable. Why they do this, I can't say.

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/prometheus-support/265)

### DIFF
--- a/config/federation/grafana/dashboards/Alert_ParserDailyVolumeTooLow.json
+++ b/config/federation/grafana/dashboards/Alert_ParserDailyVolumeTooLow.json
@@ -2,6 +2,7 @@
   "annotations": {
     "list": [
       {
+        "$$hashKey": "object:81",
         "builtIn": 1,
         "datasource": "-- Grafana --",
         "enable": true,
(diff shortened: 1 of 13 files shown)


nkinkade:
This should resolve #263.

---

Review status: 0 of 1 LGTMs obtained

---



*Comments from [Reviewable](https://reviewable.io/reviews/m-lab/prometheus-support/265#-:-LG2aiuFAArRyaqjk7RB:b-c7l336)*

stephen-soltesz:
:lgtm: -- Thank you so much for taking the time and effort to perform this migration!

One question and one minor request.

---

Review status: :shipit: complete! 1 of 1 LGTMs obtained

---

*[config/federation/prometheus/alerts.yml, line 300 at r1](https://reviewable.io/reviews/m-lab/prometheus-support/265#-LG662cKEty6OoqebgK_:-LG662cKEty6OoqebgKa:b6rwd38) ([raw file](https://github.com/m-lab/prometheus-support/blob/a4ccfdc284356249987642b1d10f86144e8ae952/config/federation/prometheus/alerts.yml#L300)):*
> ```YAML
>     prometheus_targets = "https://prometheus.mlab-oti.measurementlab.net/targets",
>     gcsbucket = "https://console.cloud.google.com/storage/browser/operator-mlab-oti/prometheus/snmp-targets",
>     dashboard = "https://grafana.mlab-oti.measurementlab.net/d/SuqnZ6Hiz/"
> ```

Don't do this but I'm wondering if we can choose this random string in advance or is it assigned by the system?

Could we adopt a convention like "staticNNN" to name static dashboards?

---

*[config/federation/prometheus/alerts.yml, line 582 at r1](https://reviewable.io/reviews/m-lab/prometheus-support/265#-LG65LBo0QR_ktrr-YEB:-LG65LBp0pQFD0fVsNAs:b9vb5nv) ([raw file](https://github.com/m-lab/prometheus-support/blob/a4ccfdc284356249987642b1d10f86144e8ae952/config/federation/prometheus/alerts.yml#L582)):*
> ```YAML
>     summary = "Today's test volume is less than 70% of nominal daily volume.",
>     hints = "Are machines online? Is data being collected? Is the parser working?",
>     url = "https://grafana.mlab-oti.measurementlab.net/d/PKqnWeNmz/"
> ```

It looks like many of the other alerts use the key name "dashboard" for this link. Would you please update this one to say `dashboard =`?

---


*Comments from [Reviewable](https://reviewable.io/reviews/m-lab/prometheus-support/265#-:-LG65jbm-4_TJcCl1LVK:b-w3luop)*

nkinkade:
Review status: :shipit: complete! 1 of 1 LGTMs obtained

---

*[config/federation/prometheus/alerts.yml, line 300 at r1](https://reviewable.io/reviews/m-lab/prometheus-support/265#-LG662cKEty6OoqebgK_:-LG67svJ4whYl6gjvr9H:bixfn4r) ([raw file](https://github.com/m-lab/prometheus-support/blob/a4ccfdc284356249987642b1d10f86144e8ae952/config/federation/prometheus/alerts.yml#L300)):*
The unique string gets [assigned automatically by the system, if not provided](http://docs.grafana.org/http_api/dashboard/#identifier-id-vs-unique-identifier-uid):

> It’s automatically generated if not provided when creating a dashboard.

I'm fine with adopting some naming convention for UIDs. I'm also fine with just letting the system create them. It would be less work to just use the ID Grafana auto-generates. What is your use case for `staticNNN`?

 It appears that our use for creating new static dashboard will be something like:

1. Create a new dashboard in the Grafana Web-UI in mlab-sandbox.
2. When happy with the dashboard, click "View JSON".
    * The JSON will contain an `id` field, which must be manually removed.
    * The JSON will contain an auto-generated `uid` field/value, which we can either leave alone, or we can change to suit some standard we have.
3. We paste the JSON into a file in the prometheus-support repo, and do the usual code review process.
4. In the event that we chose to modify the `uid` created by the system in mlab-sandbox for the new JSON dashaboard, once the new JSON dashboard is provisioned in mlab-sandbox, we will need to manually delete the original dashboard we used to model the new JSON dashboard.

---

*[config/federation/prometheus/alerts.yml, line 582 at r1](https://reviewable.io/reviews/m-lab/prometheus-support/265#-LG65LBo0QR_ktrr-YEB:-LG6ARWxCTgeKXS9mMS2:b-896fix) ([raw file](https://github.com/m-lab/prometheus-support/blob/a4ccfdc284356249987642b1d10f86144e8ae952/config/federation/prometheus/alerts.yml#L582)):*

It looks like many of the other alerts use the key name "dashboard" for this link. Would you please update this one to say `dashboard =`?
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io/reviews/m-lab/prometheus-support/265)*

PR #305, opened 2018-09-07
Adds a new SwitchDownAtSite alert

nkinkade:
Also modifies `SnmpScrapingDownAtSite` to only fire when the switch is pingable, and only when the condition exists for 6+ hours.

Currently, we get way too many `SnmpScrapingDownAtSite` alerts for transient issues that resolve themselves in the course of an hour or two. With these changes, if a switch isn't pingable (and likely down) then there is no need to fire an alert about not being able to scrape SNMP metrics. Likewise, if both SNMP scraping is down and the switch isn't pingable for 24h, then we likely need to intervene with the site host or the transit provider.

### DIFF
--- a/config/federation/prometheus/alerts.yml
+++ b/config/federation/prometheus/alerts.yml
@@ -134,6 +134,28 @@ ALERT ScraperCollectorMissingFromScraperSync
     description = "",
   }
 
+
+# SwitchSLO
+#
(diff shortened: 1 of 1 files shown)


pboothe:
It looks good to me, and I would be happy to keep the (now suppressed) alert at 2h instead of moving it up to 6h.

Issue #317, opened 2018-09-26
Prometheus 2.x web lifecycle hook times out


jheretic:
The `/-/reload` URL that prompts Prometheus to reload its configuration files times out with a `504` error when called via the public URL. This doesn't seem to affect the configuration files actually being reloaded, and does not timeout when called via the private network (meaning it's probably a timeout of the public ingress), but it would probably be a good idea to change the configmap-reload configuration so that it doesn't call through the public ingress so that errors don't show up in its log.

jheretic:
Furthermore, this causes Travis to hang when the `apply-cluster.sh` script runs and tries to hit the reload hook. We should look at how to fix and/or update that process as well, if it is even still necessary with the addition of configmap-reload.

stephen-soltesz:
This may work for prometheus-federation also: https://github.com/m-lab/prometheus-support/blob/master/k8s/data-processing-cluster/deployments/prometheus.yml#L94

nkinkade:
PR #328 should fix this. Closing.

PR #328, opened 2018-10-02
Changes the configmap-reload container to use the internal service URL.

nkinkade:
Previously it was using the public URL, and for reasons that we don't understand yet was receiving a 504 Gateway Timeout error. Apparently using the internal service URL gets around this issue.

It also removes some obsolete commented-out code that previously reloaded Prom through the public URL. This isn't necessary now that we have the configmap-reload sidecar container, which does this automatically.

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/prometheus-support/328)


nkinkade:
This PR should resolve #317.

PR #340, opened 2018-10-19
Set mlab-staging RAM to equal mlab-oti RAM

pboothe:
OOM killer killed Prometheus in staging. Prometheus in staging also is set to have a much lower RAM quota. I feel like this is not a coincidence.

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/prometheus-support/340)


pboothe:
Comments from chat: Please bump legacy RAM. 
Me: Done.

Also done: Making the staging Prometheus node-pool be highmem.

PR #353, opened 2018-10-25
Add byte-rate panel

gfr10598:
This introduces a panel that shows the submission rate, in bytes/hour, and the average filesize in bytes.

There is also now a file count metric, which I will add in a future PR.

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/prometheus-support/353)


gfr10598:
Ah - actually I already included the file rate panel too.

PR #362, opened 2018-11-06
snmp_exporter w/ static outbound IP

jheretic:
This pull request adds the Prometheus snmp_exporter as a Kubernetes deployment. It also includes a gcp-service-discovery sidecar to automatically pull down the SNMP config from the switch-config bucket, and a configmap-reload sidecar to watch for changes and reload the exporter.

In order to facilitate static outbound IPs for switch whitelisting, it is necessary to create a new nodepool and assign static IPs to it. This pull request also includes a kubeIP deployment (which runs separately on the main nodepool) to monitor the nodes of the static nodepool and assign them IPs. Because setting up the nodepool and the necessary credentials for kubeip is outside the scope of Kubernetes, instructions on doing so via gcloud have been added to the README.

As currently running, before switching over sandbox to use this instance of snmp_exporter instead of the standalone VM, the IPs `35.224.169.63` and `35.226.122.118` will need to be added to the switch whitelists. Two additional IPs each will need to be allocated for the staging and oti prometheus-federation clusters, using the directions in the README.

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/prometheus-support/362)

### DIFF
--- a/README.md
+++ b/README.md
@@ -617,6 +617,59 @@ Delete the configmaps.
 
     kubectl delete configmap blackbox-config
 
+## SNMP Exporter
+
+The SNMP exporter allows probes of SNMP endpoints, mostly switches. However,
+the switches whitelist SNMP access by IP, so we need to do some special setup.
+We're going to create a two-node pool, make sure only services that need static
+outbound IPs run on it, and then use a service called [kubeIP](https://github.com/doitintl/kubeip)
+to assign static IPs to its nodes.
+
+## Create pool
+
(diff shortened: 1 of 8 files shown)


jheretic:
Just as an additional note, I have tested to make sure that traffic originating from the respective pods is coming from the IPs specified. Also, this deployment includes two replicas for failover purposes as suggested.

jheretic:
@nkinkade no worries, between finishing a big project and working on a session proposal for IFF I didn't have time to integrate the changes yet anyway :-)

PR #428, opened 2019-04-05
Adds a 'machine' label to all metrics scraped from the platform-cluster

nkinkade:
The legacy platform uses the label `machine` in a lot of places. The new platform knows about `nodes` (in most cases), not "machines". This PR adds a `machine` label to all metrics scraped from the platform-cluster so that existing queries will still work as expected.

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/prometheus-support/428)


nkinkade:
Already implemented elsewhere. Closing.

PR #478, opened 2019-06-06
Remove static VMs target files

stephen-soltesz:
With the update of mlabconfig.py to source siteinfo API -- https://github.com/m-lab/operator/pull/284 we can now declare cloud sites in our configs along side physical sites. And, our target generation script in operator automatically includes (or excludes) services on cloud VMs.

This change removes the static VM target files from prometheus-support.

NOTE: After merge the json files must be MANUALLY removed from the project prometheus-federation cluster in the prometheus pod persistent volume. For example:

```
kubectl exec -it prometheus-server-<pod> -c prometheus sh
cd /
rm -f */vms_*
```

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/prometheus-support/478)


stephen-soltesz:
https://github.com/m-lab/dev-tracker/issues/355

PR #525, opened 2019-08-15
Improves and fixes K8s:WorkloadOverview dashboard + fixes RamBelowExpected alert

nkinkade:
Several recording rules were changed which drive the K8s:WorkloadOverview dashboard. This PR updates the panels to use the updated and renamed recording rules. It also removes a couple of unused panels, and improves the "Pod container restarts" panel to:
* Not drag down the entire dashboard
* Display more useful values. It now only displays containers where the restart count is > 10

This PR also fixes the `Hardware_RamBelowExpected` alert to not alert on cloud or testings nodes, both of which might be below the 16GB threshold.

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/prometheus-support/525)

### DIFF
--- a/config/federation/grafana/dashboards/K8s_WorkloadOverview.json
+++ b/config/federation/grafana/dashboards/K8s_WorkloadOverview.json
@@ -16,7 +16,7 @@
   "gnetId": null,
   "graphTooltip": 0,
   "id": 261,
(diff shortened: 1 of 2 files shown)


stephen-soltesz:
https://github.com/m-lab/dev-tracker/issues/315

PR #547, opened 2019-09-19
Adds an alert to notify us if a DaemonSet rollout is stuck

nkinkade:
This PR is meant to resolve this issue: https://github.com/m-lab/k8s-support/issues/281

This PR conservatively considers that if at least 2 pods have not updated in the last hour that something is wrong. The alert looks for the `delta()` of `kube_daemonset_updated_number_scheduled` over the last hour and if the number is less than 2 for 5m, an alert will fire. The reasoning for the small toleration window (5m) is that the query is already looking back over the past hour. So by the time the query is true, the issue has already been present for an hour. The alert will then only tolerate this condition for ~5 scrape cycles after that.

I've requested a review of this PR from @gfr10598, but want to also flag @pboothe and @robertodauria, in case either of them has any concerns about the configuration of this alert (e.g., maybe it is too conservative?).

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/prometheus-support/547)

### DIFF
--- a/config/federation/prometheus/alerts.yml
+++ b/config/federation/prometheus/alerts.yml
@@ -1284,6 +1284,28 @@ groups:
         pod is scheduled on. Check the status of the pod itself, if it exists.
       dashboard: https://grafana.mlab-staging.measurementlab.net/d/rJ7z2Suik/k8s-site-overview
 
+  # A DaemonSet rollout is progressing too slowly.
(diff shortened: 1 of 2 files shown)


pboothe:
I think being conservative about rollout alerts is a good thing. If a rollout is critical, we will be watching it, and if it is not critical then I really don't want alert spam from it.

PR #557, opened 2019-10-28
Two small Grafana dashboard fixes

nkinkade:
* k8s:MasterCluster: Fixes the CPU panel to give a more accurate representation of %CPU used. Currently, it sums the CPU usage across all 4 CPUs and uses that number as a % of CPU usage. This is misleading because it will give an exaggerated view of how much CPU is actually used. For example, if each CPU is 20% over a 5m average, the current panel will show an 80% CPU usage, when in actuality the usage across all 4 is only about 20%.
* k8s:WorkloadOverview: Removes an unneeded, bulky legend from the aggregate CPU usage panel.

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/prometheus-support/557)

### DIFF
--- a/config/federation/grafana/dashboards/K8s_MasterCluster.json
+++ b/config/federation/grafana/dashboards/K8s_MasterCluster.json
@@ -15,8 +15,8 @@
   "editable": true,
   "gnetId": null,
   "graphTooltip": 0,
-  "id": 63,
-  "iteration": 1564687572433,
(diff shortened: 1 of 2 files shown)


nkinkade:
@stephen-soltesz: Since @robertodauria is out sick today, would you be able to review this small PR with a couple of minor dashboard fixes?

PR #568, opened 2019-11-11
Remove unused bqx queries

stephen-soltesz:
This change removes bigquery exporter queries that we no longer use.

Four queries "bq_ndt_tests.sql", "bq_ndt_annotation.sql", "bq_gardener_parse_time.sql", "bq_ndt_server.sql" should be updated to use the new ndt5, tcpinfo and/or traceroute tables.

Only the gardener parse times and ndt_annotation rates have alerts that depend on them.

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/prometheus-support/568)


stephen-soltesz:

PR #577, opened 2019-11-20
First stab at removing all legacy PLC things

nkinkade:
Now that we have migrated all nodes to the new k8s platform clusters, we should be safe to remove all legacy alerts, recording rules, Travis build steps, etc. This PR attempts to do this, but it is a first stab and it's likely that I have missed some things or failed to take something into account.

Some things of note:
* GMX now includes both `machine` and `node` labels, so this PR removes all uses of `label_replace` which were used only to add a `node` label to its metrics.
* The neubot IPv6 Prometheus targets were being placed in the wrong directly, which was out of alignment with where gcs-downloader was looking for the targets files, so they weren't being scraped.

**NOTE**: Once we are satisfied with this PR, we will need to manually delete old, unused legacy Prom target files from GCS. We will also likely need to remove any legacy target files from the running Prometheus pods.

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/prometheus-support/577)


nkinkade:
@stephen-soltesz: I'm commenting here in Github instead of Reviewable.io because of the strange difference in comment placement, with Github being correct and Reviewable showing the comments in totally wrong places... at least wrong in comparison to where you _intended_ for them to show up.

* `LameDuckMetricMissingForNode`: there is currently no replacement for this. On the platform cluster lame-duck mode is set by a taint. When the taint exist kube-state-metrics returns a metric for the taint. When the taint  doesn't exist, no metric will exist. Because of this, I'm not sure that we can replicate this alert on the platform cluster.
* `VdlimitTooMuchUsedDisk`: I believe this one is sufficiently, at least for now, covered by the alert `PlatformCluster_DataPartitionTooFull`. Do you agree?
* `absent()` alerts: I'm all for moving certain alerts to the platform cluster instance of Prometheus to avoid prometheus-federation having to scrape the platform-cluster for every metric we might want to alert on. I'll add an agenda item to the Prometheus Sync notes document for next time.

k8s/prometheus-federation/deployments/prometheus.yml
```diff
@@ -154,8 +154,6 @@ spec:
                 "--http-source=https://storage.googleapis.com/operator-{{GCLOUD_PROJECT}}/prometheus/snmp-targets/snmpexporter.json",
                 "--http-target=/targets/script-targets/ndt_e2e.json",
                 "--http-source=https://storage.googleapis.com/operator-{{GCLOUD_PROJECT}}/prometheus/script-targets/ndt_e2e.json",
-                "--http-target=/targets/script-targets/ndt_queue.json",
```
stephen-soltesz: Please also remove the targets for legacy node-exporter, ssh806, rsyncd, mobiperf, and any others now obsolete.

stephen-soltesz:
Your replies re: specific comments in the alerts all make sense.

k8s/prometheus-federation/deployments/prometheus.yml
nkinkade: Done.

PR #596, opened 2019-12-09
Update Prometheus: Self Monitoring dashboard

stephen-soltesz:
This change removes support for the 1.x versions of Prometheus from the dashboard. And, clearer calculations for estimated "consumption rate" and "usage estimate" panels. As well, this change updates the version of prometheus & node-exporter in the data processing cluster configs to support subqueries and the same bytes available metric as other clusters.

Dashboard should be visible here: https://grafana.mlab-sandbox.measurementlab.net/d/sVklmeHik/prometheus-self-monitoring?orgId=1&var-datasource=Platform%20Cluster%20(mlab-staging)&from=now-7d&to=now

NB: the downloader cluster uses prometheus v1.6 -- this change breaks support for that configuration. That cluster should be retired in favor of migration to the new data-processing cluster.

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/prometheus-support/596)

### DIFF
--- a/config/federation/grafana/dashboards/Prometheus_SelfMonitoring.json
+++ b/config/federation/grafana/dashboards/Prometheus_SelfMonitoring.json
@@ -1,25 +1,4 @@
 {
-  "__inputs": [],
-  "__requires": [
-    {
-      "type": "grafana",
-      "id": "grafana",
-      "name": "Grafana",
-      "version": "5.1.4"
-    },
-    {
-      "type": "panel",
-      "id": "graph",
-      "name": "Graph",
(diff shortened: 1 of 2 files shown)


stephen-soltesz:
Note: that until we deploy https://github.com/m-lab/k8s-support/pull/333 to production, the filesystem panels will not work for the prod platform cluster.

stephen-soltesz:
Note also: the silly bumps in staging are due to the disk increase -- once things reach steady state the values will make much more sense.

stephen-soltesz:

PR #600, opened 2019-12-11
Update bigquery exporter version and bq_ndt_s2c query

stephen-soltesz:
This change includes a fix to the bq_ndt_s2c query and an upgrade of the bqx container version.

The bq_ndt_s2c query used a `DECLARE` statement which evidently isn't supported by default by either the older or newer versions of bqx. This change replaces is it with a function instead.

The version of bqx is updated to v1.0.0. This includes flag changes and use of the canonical port.

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/prometheus-support/600)


stephen-soltesz:

Issue #607, opened 2020-01-10
The machine label should always match the node label, if it exists


nkinkade:
Currently, when Prometheus does a federated scrape of the platform cluster instance of Prometheus the label values for `machine` and `node` may not be the same. We do not want these labels to differ because in many queries we use them interchangeably for legacy reasons.

As an example, if you query Prometheus for the following, assuming there actually are any matching taints:

`kube_node_spec_taint{cluster="platform-cluster"}`

... you will find that `machine="prometheus-platform-cluster"` and `node` will be equal to some M-Lab node e.g., mlab4.den04.measurement-lab.org. This happens for certain metrics because in the platform cluster Prometheus configs the [`machine` label is created from the label `__meta_kubernetes_pod_node_name`](https://github.com/m-lab/k8s-support/blob/master/config/prometheus/prometheus.yml#L130). In the case where the metric was discovered from Prometheus automatic service discovery and was fetched from Prometheus, then then `__meta_kubernetes_pod_node_name` will be "prometheus-platform-cluster", which isn't actually meaningful to us in most cases. 

When scraping federated metrics from the platform cluster we should probably always overwrite the `machine` label with the `node` label, if the `node` label exists.

nkinkade:
This was resolved by #608. Closing.

PR #608, opened 2020-01-10
Set the 'machine' label equal to the 'node' label, if the label 'node' label exists.

nkinkade:
This PR is intended to resolve issue #607.

In summary, in the case where a metric has a `node` label, then the `machine` label should be overwritten with the `node` label's value.

NOTE: I've attempted to spot check this for sanity, and I _believe_ it is the correct thing to do and safe, but a second set of eyes will be good. Hence this  PR, I guess.

An an indirect way this PR is part of PR https://github.com/m-lab/mlab-ns/pull/216, which attempts to resolve https://github.com/m-lab/mlab-ns/issues/215

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/prometheus-support/608)

### DIFF
--- a/config/federation/prometheus/prometheus.yml.template
+++ b/config/federation/prometheus/prometheus.yml.template
@@ -353,6 +353,21 @@ scrape_configs:
         target_label: experiment
         replacement: ndt.iupui
 
+      # Rewrite the machine label to make it easier to join these metrics with
+      # existing metrics that use machine instead of the node label.
+      #
+      # NOTE: Relabeling that happens in the Prometheus instance in the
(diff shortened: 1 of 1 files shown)


nkinkade:
@stephen-soltesz: I believe now that this PR may not be working as I was intending in all cases. For example, in mlab-sandbox the query:
```
lame_duck_experiment{cluster="platform-cluster"}
```
Does not have a `machine` label, but in staging it does. I suspect this is because when a `node` label does not exist Prometheus interprets the rule I added as "then remove any existing `machine` label."

Please hold off on this PR until I can figure out a better way to handle this.

PR #615, opened 2020-01-16
Add support for automatically removing datasources

stephen-soltesz:
This change adds a new file `deleted.yml` that uses the `deletedDatasources` directive as described in https://grafana.com/docs/grafana/latest/administration/provisioning/#datasources

This change requires eventually removing the records in `deletedDatasources` but seems to be a no-op until we do.

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/prometheus-support/615)


stephen-soltesz:

PR #636, opened 2020-03-06
Increases threshold for platform cluster scrape job failing/absent.

nkinkade:
The threshold is being increased from 4m to 10m. This PR also increases the threshold for missing workloads to 15m so that a failing/absent scrape job doesn't allow a bunch of missing DaemonSet alerts to fire (they are inhibited by Alertmanager).

10m seems like a bit longer than I'd like to allow for the platform cluster scrape job to down, but I'm not sure what else to do here.

This is a workaround (fix?) for m-lab/k8s-support/issues/371.

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/prometheus-support/636)

### DIFF
--- a/config/federation/prometheus/alerts.yml
+++ b/config/federation/prometheus/alerts.yml
@@ -949,7 +949,7 @@ groups:
 # time for this to be down before someone knows about it.
   - alert: PlatformCluster_FederationScrapeJobFailing
     expr: up{job="platform-cluster"} == 0
-    for: 4m
+    for: 10m
     labels:
       repo: ops-tracker
       severity: ticket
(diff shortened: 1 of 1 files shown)


nkinkade:
I'm not thrilled about extending the alert threshold as a way around this, but I don't know what we can do about it right now. Best I can tell, slow startup times [are a known issue](https://github.com/prometheus/prometheus/issues/6526) for busy Prometheus instances.

stephen-soltesz:
I've day dreamed about experimenting with alternate storage / write backends that might help. But, I don't expect to have time to investigate those any time soon. :(

PR #637, opened 2020-03-09
Only report alerts from the last week

stephen-soltesz:
This change is [Reviewable](https://reviewable.io/reviews/m-lab/prometheus-support/637)


stephen-soltesz:
My original thinking about the utility of that panel is that it would show "what alerts fired in the last week" -- We can check AM directly to see what alerts are currently firing. Knowing what fired in the last week should help spot alerts that fire too frequently or when we have just too many firing per week.

nkinkade:
Then maybe an improvement would be to filter on `status="firing"` and just ignore the resolved ones?

PR #654, opened 2020-04-21
Change downloader alert wait threshold

stephen-soltesz:
After downloader restarts and before a successful download, the LastSuccessTime metric is initialized to zero and the alert using that metric can fire when it does not need to.

A pod restart can occur as a result of normal k8s node restarts and pod migration between nodes.

1h should allow enough time for the download to attempt and succeed after a restart.

Fixes https://github.com/m-lab/dev-tracker/issues/571

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/prometheus-support/654)


stephen-soltesz:
https://github.com/m-lab/dev-tracker/issues/580

PR #658, opened 2020-04-27
Add script-exporter to federation cluster with access-token e2e monitoring

stephen-soltesz:
This PR adds several changes to the prometheus federation cluster to support access-token enabled e2e monitoring using the locate service.

This change adds a new script exporter deployment with configuration to target ndt5 servers using the ndt5-client. When ndt5-client runs it is invoked by the monitoring-token command, which uses the LOCATE_URL and MONITOR_SIGNING_KEY from the environment to request a valid monitoring token from the locate service in the current project. (projects mlab-oti/mlab-ns are not yet supported).

We are adding a second script exporter. Normally target files are read from a directory `/script-targets/*.json`. Because the config also targets a static script exporter, this PR adds a second directory `/script-targets-federation/*.json` and script exporter target within the federation cluster.

The script exporter is assigned to the static "outbound-ip" node pool so that the set of source IPs for the e2e tests remains small and static.

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/prometheus-support/658)

### DIFF
--- a/apply-global-prometheus.sh
+++ b/apply-global-prometheus.sh
@@ -180,6 +180,11 @@ if [[ "${PROJECT}" != "mlab-oti" ]] ; then
   kubectl create secret generic script-exporter-secret \
     "--from-file=/tmp/monitoring-signer-key.json" \
     --dry-run -o json | kubectl apply -f -
+
+else
+  # NOTE: disable for mlab-oti until we're deploying there also.
+  mv k8s/prometheus-federation/deployments/script-exporter.yml \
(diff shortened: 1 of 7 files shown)


stephen-soltesz:
https://github.com/m-lab/dev-tracker/issues/574

stephen-soltesz:
Looks like ndt-server needs to be tagged and updated in k8s-support for the metrics to be labeled properly. This change should still as-is until that change is in place.

stephen-soltesz:
This is not ready for review yet.

PR #684, opened 2020-05-19
Update sx to include random mtime change

stephen-soltesz:
This change updates the script-exporter image built by the script-exporter-support repo to include recent improvements to the cache mtime randomization to be resilient to accidental sychronization. The result of that change is that the expected frequency of ndt measurements is every 5min rather than 10min.

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/prometheus-support/684)


stephen-soltesz:
https://github.com/m-lab/dev-tracker/issues/574

PR #703, opened 2020-06-16
Add basic bq billing information

stephen-soltesz:
This change introduces a new bigquery-exporter query to report service-level aggregate GCP billing and credit usage.

Because data exported to bigquery may be delayed for up to 1.6days (by observation) and because prometheus can only report "current" values, this change exports the hourly billing from two day prior to the current hour.

As well, this query uses a manually created view that unifies the staging and oti tables. There may be permission adjustments necessary to the default GKE user so that it can query that view in staging and oti projects.

I am also considering one or two additional queries that would aggregate by day, and another that would export only by GCE VMs, because there is machine type information that could help identify unnecessary costs.

This change is [Reviewable](https://reviewable.io/reviews/m-lab/prometheus-support/703)

### DIFF
--- a/config/federation/bigquery/bq_billing_hourly.sql
+++ b/config/federation/bigquery/bq_billing_hourly.sql
@@ -29,7 +29,7 @@ WITH recent_billing AS (
 , net_hourly_billing AS (
   SELECT
     CONCAT(service.description, ":", sku.description) AS service,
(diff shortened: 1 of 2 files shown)


stephen-soltesz:
https://github.com/m-lab/dev-tracker/issues/603

PR #706, opened 2020-06-18
Update recording rules & global dashboard with ndt7 metrics

stephen-soltesz:
The Global Test rate dashboard is self-contained.

The recording rules make it possible to additionally update the Early Warning dashboard with ndt7 metrics once the new metrics are available.

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/prometheus-support/706)

PR #707, opened 2020-06-19
Update NDT:EarlyWarning with ndt7 metrics

stephen-soltesz:
This change updates the NDT Early Warning dashboard to include ndt7 metrics.

The recording rules for ndt7 metrics were only recently added in https://github.com/m-lab/prometheus-support/pull/706 as a result most of current historical data does not include these. This means that sum operations `+` would return no metrics until there were both ndt5 and ndt7 metrics available. To work around this, and provide an easy way to clean up in the future, every ndt5+ndt7 query has a complementary query that matches a pattern like:

```
1st query: (ndt5 + ndt7)
2nd query: ndt5 unless ndt7
```

These two queries provide timeseries before (2nd) and after (1st) the ndt7 metrics are available. As well, the "unless" queries can be removed all at once in the future easily.

As well, this change removes some legacy metrics that are no longer relevant (inotify, vdlimit).

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/prometheus-support/707)

### DIFF
--- a/config/federation/grafana/dashboards/NDT_EarlyWarning.json
+++ b/config/federation/grafana/dashboards/NDT_EarlyWarning.json
@@ -15,11 +15,13 @@
   "editable": true,
   "gnetId": null,
   "graphTooltip": 0,
-  "iteration": 1564682328114,
+  "id": 338,
+  "iteration": 1592574816804,
   "links": [],
   "panels": [
     {
       "content": "# Real time",
+      "datasource": null,
       "gridPos": {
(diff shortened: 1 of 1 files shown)


stephen-soltesz:
The travis build passed - but the ci check isn't showing up in github -- travis has been acting strangely lately. Merging anyway.

stephen-soltesz:
https://github.com/m-lab/dev-tracker/issues/603

PR #765, opened 2020-12-08
Add a high-level summary dashboard

stephen-soltesz:
This is a start to a high-level summary dashboard that verifies that the platform is still running end to end.

The initial set of panels include aggregate network usage, total ndt measurements, bytes uploaded to GCS, and rows processed by the batch ETL pipeline.

The panels are oriented left to right to mirror the path of data through the system.

Possibilities for future refinements include:
* request rates to Locate API
* metrics from the ETL pipeline corresponding only to the current day rather than batch
* 24h summaries using recording rules for faster loading
* BQ metrics like bytes queries, number of queries, or users accessing our data.

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/prometheus-support/765)

### DIFF
--- a/config/federation/grafana/dashboards/KeepThePlatformRunning.json
+++ b/config/federation/grafana/dashboards/KeepThePlatformRunning.json
@@ -15,8 +15,7 @@
   "editable": true,
   "gnetId": null,
   "graphTooltip": 1,
-  "id": 350,
-  "iteration": 1607291390785,
+  "iteration": 1607464932567,
   "links": [],
   "panels": [
     {
@@ -58,12 +57,12 @@
       "seriesOverrides": [
         {
           "alias": "Current",
-          "color": "#56A64B",
+          "color": "#37872D",
           "linewidth": 2
         },
         {
           "alias": "One Week",
-          "color": "#56A64B"
+          "color": "rgb(58, 99, 51)"
         },
         {
           "alias": "Two Weeks",
@@ -76,16 +75,19 @@
       "targets": [
         {
           "expr": "sum(rate(ifHCOutOctets{ifAlias=\"uplink\"}[5m]))",
+          "intervalFactor": 2,
           "legendFormat": "Current",
           "refId": "C"
         },
         {
           "expr": "sum(rate(ifHCOutOctets{ifAlias=\"uplink\"}[5m] offset 7d))",
+          "hide": true,
           "legendFormat": "One Week",
           "refId": "A"
         },
         {
           "expr": "sum(rate(ifHCOutOctets{ifAlias=\"uplink\"}[5m] offset 14d))",
+          "hide": true,
           "legendFormat": "Two Weeks",
           "refId": "B"
         }
@@ -94,7 +96,7 @@
       "timeFrom": null,
       "timeRegions": [],
       "timeShift": null,
-      "title": "Aggregate network usage (bits/s)",
+      "title": "Switch: Global Network Utilization (outbound)",
       "tooltip": {
         "shared": true,
         "sort": 2,
@@ -112,7 +114,7 @@
         {
           "decimals": null,
           "format": "bps",
-          "label": null,
+          "label": "Bits / Sec",
           "logBase": 1,
           "max": null,
           "min": null,
@@ -153,7 +155,7 @@
         "current": false,
         "max": false,
         "min": false,
-        "show": true,
+        "show": false,
         "total": false,
         "values": false
       },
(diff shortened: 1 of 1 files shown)


nkinkade (commented):
A few clarifying questions and suggestions:

Aggregate bandwidth panel
* I was interested to see that the aggregate bandwidth reported by the switches is significantly lower than the [aggregate bandwidth reported by containers](https://grafana.mlab-sandbox.measurementlab.net/d/tZHLFQRZk/k8s-workload-overview?orgId=1&refresh=5m&fullscreen&panelId=30&from=now-7d&to=now&edit) (cAdvisor), by a factor of more than 2. Is there something obvious that I'm missing that would explain this? Maybe the container stats are double counting bytes somewhere?
* This was likely intentional on your part, but the colors of the lines are close enough to one another that I find it a little hard to differentiate them. Some more contrasty colors would be helpful, at least to me.

NDT Global Test Rate
* Could you modify the title to indicate that we are looking at tests per minute? I only say this because generally `rate()` in Prometheus is per second, but I see you are multiplying the rate by 60. It would just be helpful to know at a glance, without having to look at the query, that we are seeing tests per minute.
* Same thing about similarly colored lines on the graph.

Successful Bytes Uploaded
* It's not clear from the title alone where the bytes came from and where they were uploaded to. Maybe change the title to something like `Pusher: successful bytes uploaded`?
* pcap and tcpinfo dwarf all other bytes by so many that the others may as well not even exist on the panel. I don't know what the solution could be, but it would be nice if we could actually see the others to look for variations. Maybe move pcap and tcp-info to the right-Y, so that each Y can scale appropriately?

Total Rows Written
* Similar to others, is there a way to make the title more descriptive of what was written where?
* Just for my own information, I see sidestream in the list of "legacy" table writes. Is this be cause of continual reprocessing of old data.

> *[Reviewable](https://reviewable.io/reviews/m-lab/prometheus-support/765#-:-MO342P-7-KCLgWpxAMg:b3syz9t)* status: 0 of 1 approvals obtained (waiting on @nkinkade)

stephen-soltesz (commented):
Thank you for this excellent feedback.

I think I noticed the same thing from the cadvisor metrics and used the switch metrics directly instead. Could this be due to container bytes being also counted by the host?

I've normalized the units to "X / min" for all panels other than the "switch", b/c everyone thinks in terms of bits/sec for network rates.

I've updated the panel titles to be more descriptive.

I've simplified what is plotted to a single line wherever possible. The ETL panel is reading from two datasources, so that is reduced to simply "legacy" and "standard".

Yes, sidestream is reprocessed as part of the continuous reprocessing of older data.

> *[Reviewable](https://reviewable.io/reviews/m-lab/prometheus-support/765#-:-MO3TidM3NygsytzMTOM:br8o518)* status: 0 of 1 approvals obtained

stephen-soltesz (commented):
PTAL?

> *[Reviewable](https://reviewable.io/reviews/m-lab/prometheus-support/765#-:-MO3Upj89tfvFxry4JeS:b-833305)* status: 0 of 1 approvals obtained

nkinkade (approved):
:lgtm:

Thank you. The panel titles and Y descriptions make these panels much more evident to me now.

For outbound traffic, [the recording rule for the container network transmitted data is _supposed_ to exclude all pods that have `hostNetwork=true`](https://github.com/m-lab/k8s-support/blob/master/config/prometheus/rules.yml#L117). So I would think it is only recording NDT network bytes a single time. I did a test here for ord05 with the query that makes up the recording rule and the plots are very close between disco and container metrics:

https://grafana.mlab-sandbox.measurementlab.net/d/vKyDQejWz/kinkade-ndt-data-rates?tab=queries&orgId=1&fullscreen&panelId=2&edit&from=now-7d&to=now&var-datasource=Platform%20Cluster%20(mlab-oti)&var-site=ord05

This is probably something worth digging into deeper. I have a tendency to want to trust the switch metrics more.

Regarding having offsets of current, 7 day and 14 day, I thought that was actually fairly interesting to see. I suppose one can now edit the panel and enable week and two week offsets, if someone wanted them. My issue wasn't the three lines, but the colors of the lines. I'm nearly sure you know this, but to be 100% sure, you can change the colors of lines to whatever you want.

> *[Reviewable](https://reviewable.io/reviews/m-lab/prometheus-support/765#-:-MO3VUBY7PK2hUx3uORf:bt6ms59)* status: :shipit: complete! 1 of 1 approvals obtained

stephen-soltesz (commented):
I agree it would be better to understand whether or underwhat conditions we can use the cadvisor network traffic metrics. I also tend to trust the switch metrics more.

Yeah, I understood your comment about the colors. But, I concluded that everything was a  little too busy already and simplifying is preferable for a "global, make sure the lights are still on" overview. Your feedback helped make things clearer. :+1:

> *[Reviewable](https://reviewable.io/reviews/m-lab/prometheus-support/765#-:-MO3dj9b1ySmdlCOfM-C:b-6qedfu)* status: :shipit: complete! 1 of 1 approvals obtained

PR #767, opened 2020-12-11
Change BMC_CredentialsNotFound to only fire after 1h

robertodauria:
The E2E test results are cached for an hour. If credentials are created and then the site is immediately taken out of GMX, this alert would fire because credentials weren't there when the last check happened, then clear itself at the next check.

By requiring that the condition is true for `1h` before the alert fires we prevent this.

Not sure if this also closes https://github.com/m-lab/reboot-service/issues/40 as it's a workaround. @nkinkade what do you think?

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/prometheus-support/767)

### DIFF
--- a/config/federation/prometheus/alerts.yml
+++ b/config/federation/prometheus/alerts.yml
@@ -1508,7 +1508,7 @@ groups:
     expr: |
         reboot_e2e_result{status="credentials_not_found"}
         unless on(machine) gmx_machine_maintenance == 1
-    for: 5m
+    for: 1h
     labels:
       repo: ops-tracker
       severity: ticket


nkinkade (approved):
:lgtm:, but could/should we bump that to `1d` instead of `1h`. The `BMC_ConnectionFailed` alert won't fire until the condition has been true for `1d`. Is there a reason that this alert should be more sensitive?

> *[Reviewable](https://reviewable.io/reviews/m-lab/prometheus-support/767#-:-MOHn_wvDY19fZ2Kucqz:b5sr4us)* status: :shipit: complete! 1 of 1 approvals obtained (waiting on @nkinkade)

nkinkade (commented):
That makes sense. Let's try 1h and see how it goes.

> *[Reviewable](https://reviewable.io/reviews/m-lab/prometheus-support/767#-:-MOHyH90EO3J8kCJVNIr:b-1tb7i5)* status: :shipit: complete! 1 of 1 approvals obtained

PR #773, opened 2021-01-15
Limits MachineRunningWithoutK8sNode to only ssh IPv4 probes.

nkinkade:
Limits query to only use the `ssh_v4_online` probe to eliminate duplicate results for both IPv4 and IPv6 probes, making the problem seem worse than it is (by double). This should cause the query to only ever return a single result per node where the problem exists, instead of 2.

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/prometheus-support/773)


stephen-soltesz (approved):
There's a potential blind spot that this creates, but it's probably okay.

> *[Reviewable](https://reviewable.io/reviews/m-lab/prometheus-support/773#-:-MR64cxIEHD-pAzFMf8r:b-2rsfy)* status: :shipit: complete! 1 of 1 approvals obtained

PR #781, opened 2021-02-05
Add script to delete Grafana datasources

robertodauria:
Currently Grafana does not seem to be updating provisioned datasource. A workaround which we documented in this repository's README is to remove them via the Grafana APIs and then restart the pod.

Since we added an OAuth proxy in front of Grafana, there is an additional step which is grabbing a valid session cookie from a browser and pass it via curl when making requests to the API.

This PR adds a script to try to make the process (slightly) less cumbersome.

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/prometheus-support/781)

### DIFF
--- a/delete-grafana-ds.sh
+++ b/delete-grafana-ds.sh
@@ -6,11 +6,11 @@
 #    ./delete-grafana-ds.sh mlab-sandbox 42 session-id api-key
 #
 # In order to use this script you need to:
-# 1. Be logged into Grafana on a browser
-# 2. Get the value of the _oauth2_proxy cookie (the session-id)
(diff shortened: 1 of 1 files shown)


stephen-soltesz (approved):
Heavens.. :lgtm: -- if this is the less cumbersome way let's do this as infrequently as possible.

> *[Reviewable](https://reviewable.io/reviews/m-lab/prometheus-support/781#-:-MSnIyUa6_9VPy8KUsRT:b-kbzl2i)* status: :shipit: complete! 1 of 1 approvals obtained

PR #785, opened 2021-02-11
Enabled husky.neu.edu in oauth2-proxy.yml

robertodauria:
oauth2-proxy does not seem to automatically authorize subdomains, so `neu.edu` does not allow `husky.neu.edu` to log in. This change explicitly allows `husky.neu.edu`.

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/prometheus-support/785)


critzo (approved):
LGTM

PR #787, opened 2021-02-12
Adds new alert for machine uptime + only scrape annotation port

### LINKED ISSUE #786: Prometheus is erroneously scraping port 9393 of github-receiver and failing
This annotation is apparently not working:

https://github.com/m-lab/prometheus-support/blob/master/k8s/prometheus-federation/deployments/github-receiver.yml#L16

And automatic service discovery is [apparently scraping anything with a declared `containerPort` in the pod](https://github.com/m-lab/prometheus-support/blob/master/config/federation/prometheus/prometheus.yml.template#L157):

https://github.com/m-lab/prometheus-support/blob/master/k8s/prometheus-federation/deployments/github-receiver.yml#L34

Maybe we can remove the explicit `containerPort`? In any case, we need to figure out how to stop Prometheus from attempting the scrape on port 9393, and failing continually.

nkinkade:
This PR adds a new alert which fires when a machine hasn't been rebooted in more than 90d. It builds on the [recent work to cause all platform nodes to get rebooted when up for more than 60d](https://github.com/m-lab/epoxy-images/pull/200).

It also adds a new `relabel_config` config for the kubernetes-pods job which will cause the job to look for the `prometheus.io/port` annotation on a pod, and if it exists, to only scrape the pod on that port.

This PR is expected to resolve #786.

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/prometheus-support/787)

### DIFF
--- a/config/federation/prometheus/alerts.yml
+++ b/config/federation/prometheus/alerts.yml
@@ -1489,6 +1489,18 @@ groups:
         and/or system messages.
       dashboard: https://grafana.mlab-oti.measurementlab.net/d/rJ7z2Suik/k8s-site-overview
 
+  - alert: PlatformHardware_MachineUpTooLong
+    expr: |
+      (time() - node_boot_time_seconds{node=~"(master|mlab[1-4])-.+"}) / (60 * 60 * 24) > 90
+    for: 10m
+    labels:
+      repo: ops-tracker
+      severity: ticket
+    annotations:
+      summary: A machine has not been rebooted for too long
+      description: https://github.com/m-lab/ops-tracker/wiki/Alerts-&-Troubleshooting#platformhardware_machineuptimetoolong
+      dashboard: https://grafana.mlab-oti.measurementlab.net/d/_fugwnWZk/ops-tactical-and-sre-overview
+
 # BMC alerts
 # The Reboot API performs an E2E connection test on a BMC every time its
 # /v1/e2e endpoint is scraped, and reports failure via this metric.
--- a/config/federation/prometheus/alerts.yml
+++ b/config/federation/prometheus/alerts.yml
@@ -1498,7 +1498,7 @@ groups:
       severity: ticket
     annotations:
       summary: A machine has not been rebooted for too long
-      description: https://github.com/m-lab/ops-tracker/wiki/Alerts-&-Troubleshooting#platformhardware_machineuptimetoolong
+      description: https://github.com/m-lab/ops-tracker/wiki/Alerts-&-Troubleshooting#platformhardware_machineuptoolong
       dashboard: https://grafana.mlab-oti.measurementlab.net/d/_fugwnWZk/ops-tactical-and-sre-overview
 
 # BMC alerts
--- a/config/federation/prometheus/prometheus.yml.template
+++ b/config/federation/prometheus/prometheus.yml.template
@@ -185,6 +185,15 @@ scrape_configs:
         action: keep
         regex: (\d+)
 
+      # Check for the prometheus.io/port=<port> annotation. If it exists, then
+      # only scrape that one port, no matter how many containerPorts are
+      # declared.
+      - source_labels: [__address__, __meta_kubernetes_pod_annotation_prometheus_io_port]
+        action: replace
+        regex: ([^:]+)(?::\d+)?;(\d+)
+        replacement: $1:$2
+        target_label: __address__
+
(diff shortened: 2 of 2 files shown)


stephen-soltesz (approved):
A clarifying question. I have a preferred answer. I'm not sure that the other answer is blocking as long as it doesn't mess anything up with prometheus. So, :lgtm:

> *[Reviewable](https://reviewable.io/reviews/m-lab/prometheus-support/787#-:-MTNW5s74VmwzC0L4F-0:bnh92un)* status: :shipit: complete! 1 of 1 approvals obtained (waiting on @nkinkade)

___
*[config/federation/prometheus/prometheus.yml.template, line 195 at r1](https://reviewable.io/reviews/m-lab/prometheus-support/787#-MTNVhK90Va74nt5h_2W:-MTNVhK90Va74nt5h_2X:bdwuqwg) ([raw file](https://github.com/m-lab/prometheus-support/blob/596861172792b7e41e42d541912afeae36b57b20/config/federation/prometheus/prometheus.yml.template#L195)):*
> ```
>         regex: ([^:]+)(?::\d+)?;(\d+)
>         replacement: $1:$2
>         target_label: __address__
> ```

This is interesting. This is jogging my memory that maybe I wasn't able to take the port annotation further b/c I wanted to do an equality check which isn't quite possible with regexes. And, I wasn't sure what to try next. This approach is a good idea.

Since it seems that prometheus discovers both ports as separate "services", does this create two service discovery targets for the same port? Does the target get scraped twice? (I hope not, but want to be sure.)

nkinkade (commented):
> *[Reviewable](https://reviewable.io/reviews/m-lab/prometheus-support/787)* status: :shipit: complete! 1 of 1 approvals obtained (waiting on @stephen-soltesz)

___
*[config/federation/prometheus/prometheus.yml.template, line 195 at r1](https://reviewable.io/reviews/m-lab/prometheus-support/787#-MTNVhK90Va74nt5h_2W:-MTN_05R2M-dioeCkBRu:btm7tbw) ([raw file](https://github.com/m-lab/prometheus-support/blob/596861172792b7e41e42d541912afeae36b57b20/config/federation/prometheus/prometheus.yml.template#L195)):*

This is interesting. This is jogging my memory that maybe I wasn't able to take the port annotation further b/c I wanted to do an equality check which isn't quite possible with regexes. And, I wasn't sure what to try next. This approach is a good idea.

Since it seems that prometheus discovers both ports as separate "services", does this create two service discovery targets for the same port? Does the target get scraped twice? (I hope not, but want to be sure.)
</blockquote></details>

This is a good question, and one that crossed my mind as well... that Prometheus would discover two ports (9393, 9990), generate two targets, and then this rule would simply swap out 9393 for 9990, and the target would get scraped twice per interval. I tried it anyway, and it _seems_ like there is only a single target generated. When you look at the target status page for mlab-sandbox, there is only a single target listed in the kubernetes-pods job for github-receiver with a port of 9990:

https://prometheus.mlab-sandbox.measurementlab.net/targets

My assumption is, and, granted, it is just an assumption at this point, that because these relabels happen prior to scraping that Prometheus may compile targets and is smart enough to not scrape the same target spec twice in a cycle... maybe eliminating duplicates?

As far as this method of doing things. I got it from this commit:

https://github.com/prometheus/prometheus/commit/03a9e7f72e072c6d29f422425d8acd91a957836b#diff-b6b2e859b4b0b34ddc85f2d84ca52037d6c28a4a3d0b5c203d1f0728a2afcd16

... it was only after reading that commit that I realized that  `prometheus.io/.*` annotations are not some formal mechanism supported by prometheus for service discovery, but are arbitrary, and nothing more than basically indicators that operators can manually leverage to do something, or not, in the relabel phase.  The annotation could just as easily be something `rofl.io/funny_port=9990`.

nkinkade (commented):
@stephen-soltesz: Another you may or may not have noticed, but which I failed to call out in this PR, is that the new alert leverages a new ops-tracker wiki page for `Alerts & Troubleshooting`. Instead of an alert description, there is nothing, but a link to [the alert's description in the wiki](https://github.com/m-lab/ops-tracker/wiki/Alerts-&-Troubleshooting#platformhardware_machineuptimetoolong).

> *[Reviewable](https://reviewable.io/reviews/m-lab/prometheus-support/787#-:-MTNbADUCun5knP8Di3S:b-s5qdhq)* status: :shipit: complete! 1 of 1 approvals obtained (waiting on @stephen-soltesz)

PR #802, opened 2021-03-19
Adds alerts for ndt7 mlab-ns queries

nkinkade:
Currently, we have alerts for when too many ndt5 services are down, but none for ndt7, even though the vast majority the platform's traffic comes from ndt7 test. This PR attempts to rectify that issue. I have also renamed the alerts to indicate whether we are talking about ndt5 or ndt7, and uses all caps for acronyms too.

Additionally, a [new alert description has been added to the ops-tracker Alerts & Troubleshooting wiki page](https://github.com/m-lab/ops-tracker/wiki/Alerts-&-Troubleshooting#PlatformCluster_TooManyNDTServersDown) for this class of alert.

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/prometheus-support/802)

### DIFF
--- a/config/federation/grafana/dashboards/MLABNS_PrometheusQueries.json
+++ b/config/federation/grafana/dashboards/MLABNS_PrometheusQueries.json
@@ -1,0 +1,857 @@
+{
+  "annotations": {
+    "list": [
+      {
+        "builtIn": 1,
+        "datasource": "-- Grafana --",
+        "enable": true,
+        "hide": true,
+        "iconColor": "rgba(0, 211, 255, 1)",
+        "name": "Annotations & Alerts",
+        "type": "dashboard"
+      }
+    ]
+  },
+  "editable": true,
+  "gnetId": null,
+  "graphTooltip": 0,
+  "id": 357,
+  "iteration": 1616437301517,
+  "links": [],
+  "panels": [
+    {
+      "aliasColors": {},
+      "bars": false,
+      "dashLength": 10,
+      "dashes": false,
+      "datasource": "$datasource",
+      "fieldConfig": {
+        "defaults": {
+          "custom": {}
+        },
+        "overrides": []
+      },
+      "fill": 1,
+      "fillGradient": 0,
+      "gridPos": {
+        "h": 7,
+        "w": 8,
+        "x": 0,
+        "y": 0
+      },
+      "hiddenSeries": false,
+      "id": 2,
+      "legend": {
+        "avg": false,
+        "current": false,
+        "max": false,
+        "min": false,
+        "show": false,
+        "total": false,
+        "values": false
+      },
+      "lines": true,
+      "linewidth": 1,
+      "nullPointMode": "null",
+      "options": {
+        "alertThreshold": true
+      },
+      "percentage": false,
+      "pluginVersion": "7.3.5",
+      "pointradius": 2,
+      "points": false,
+      "renderer": "flot",
+      "seriesOverrides": [],
+      "spaceLength": 10,
+      "stack": false,
+      "steppedLine": false,
+      "targets": [
+        {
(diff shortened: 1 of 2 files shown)


stephen-soltesz (approved):
:lgtm: -- some questions to consider -- not necessarily blocking this PR.

> *[Reviewable](https://reviewable.io/reviews/m-lab/prometheus-support/802#-:-MWBeaKQDX0-re-Eb8BT:bu6q0p0)* status: :shipit: complete! 1 of 1 approvals obtained (waiting on @nkinkade)

___
*[config/federation/prometheus/alerts.yml, line 356 at r1](https://reviewable.io/reviews/m-lab/prometheus-support/802#-MWBetPl5ju3Cr8hhGVA:-MWBetPl5ju3Cr8hhGVB:b5zqx6c) ([raw file](https://github.com/m-lab/prometheus-support/blob/8a536ddae81a59e653b9f1995255e9df9146e162/config/federation/prometheus/alerts.yml#L356)):*
> ```YAML
>     expr: |
>       (
>         sum(
> ```

Is there anything we can do to simplify management of these very complex expressions? e.g. recording rules?

While it's different than the queries mlab-ns uses, could we combine the v4 and v6 conditions? The dashboard could differentiate which condition it is if an alert fired for the generic case. I'm thinking from the perspective of making these alerts fewer and easier to manage (somehow) without losing fidelity of signal.

Do we need both ndt5 and ndt7 alerts? Both ipv4 and ipv6 conditions? With and without TLS?

A "mega query" that asks, are all of these pieces working? if so, good, if not, then we can sort out which one when handling the alert.

This is just something to consider, not necessarily something to change now.
___
*[config/federation/prometheus/alerts.yml, line 381 at r1](https://reviewable.io/reviews/m-lab/prometheus-support/802#-MWBh6yi6FZ5bvzt12Sj:-MWBh6yi6FZ5bvzt12Sk:b-c47xnp) ([raw file](https://github.com/m-lab/prometheus-support/blob/8a536ddae81a59e653b9f1995255e9df9146e162/config/federation/prometheus/alerts.yml#L381)):*
> ```YAML
>       summary: Less than 90% of ndt7 experiments are online according to mlab-ns.
>       description: https://github.com/m-lab/ops-tracker/wiki/Alerts-&-Troubleshooting#PlatformCluster_TooManyNDTServersDown
>       dashboard: https://grafana.mlab-oti.measurementlab.net/d/tZHLFQRZk/k8s-workload-overview
> ```

You're working on a better dashboard for these alerts than the workload overview? Should that new dashboard be part of this PR? The prototype you showed me in slack is more helpful than the workload overview, imo, for these alerts.

nkinkade (commented):
> *[Reviewable](https://reviewable.io/reviews/m-lab/prometheus-support/802)* status: :shipit: complete! 1 of 1 approvals obtained (waiting on @stephen-soltesz)

___
*[config/federation/prometheus/alerts.yml, line 356 at r1](https://reviewable.io/reviews/m-lab/prometheus-support/802#-MWBetPl5ju3Cr8hhGVA:-MWQBQsd489hnKtlx7QY:b6m486l) ([raw file](https://github.com/m-lab/prometheus-support/blob/8a536ddae81a59e653b9f1995255e9df9146e162/config/federation/prometheus/alerts.yml#L356)):*

Is there anything we can do to simplify management of these very complex expressions? e.g. recording rules?

While it's different than the queries mlab-ns uses, could we combine the v4 and v6 conditions? The dashboard could differentiate which condition it is if an alert fired for the generic case. I'm thinking from the perspective of making these alerts fewer and easier to manage (somehow) without losing fidelity of signal.

Do we need both ndt5 and ndt7 alerts? Both ipv4 and ipv6 conditions? With and without TLS?

A "mega query" that asks, are all of these pieces working? if so, good, if not, then we can sort out which one when handling the alert.

This is just something to consider, not necessarily something to change now.
</blockquote></details>

I'm confident that there is _some_ way that we could simplify these alerts. One thing that occurs to me is that we could separate some of the individual `OR`d parts of the queries into their own alerts. For example, we already have an alert for when disk space is getting low on a node, so could probably just remove that entire part to simplify the queries. 

Whether we need alerts for ndt5 and ndt7, ipv4 and ipv6, tls and not, etc, I don't know. Since our alert thresholds are different for ipv4 and ipv6, it seems like those should stay distinct and separate. For ndt5 and ndt7, we could perhaps combine them into a single query which looks for the case when in, combination, they go to less than some threshold. `(ndt5_success + ndt7_successe) / (ndt5_probes + ndt7_probes)`?

For this PR I'll leave them the way they are.
___
*[config/federation/prometheus/alerts.yml, line 381 at r1](https://reviewable.io/reviews/m-lab/prometheus-support/802#-MWBh6yi6FZ5bvzt12Sj:-MWQ3UrSC3B0czSje4pj:b12ynoh) ([raw file](https://github.com/m-lab/prometheus-support/blob/8a536ddae81a59e653b9f1995255e9df9146e162/config/federation/prometheus/alerts.yml#L381)):*

You're working on a better dashboard for these alerts than the workload overview? Should that new dashboard be part of this PR? The prototype you showed me in slack is more helpful than the workload overview, imo, for these alerts.
</blockquote></details>

Good catch. I actually meant to do this. My initial intention was to add the link to the new dashboard here, with the assumption that it would make it to production in a while. But I think the dashboard is good enough for a start, so I've added it to the PR so that we don't have broken links in potential alerts, even for a week or two.

stephen-soltesz (approved):
:lgtm:

> Reviewed 2 of 2 files at r2.
> *[Reviewable](https://reviewable.io/reviews/m-lab/prometheus-support/802#-:-MWQDzO7EoAXbUdjfoNo:bnfp4nl)* status: :shipit: complete! 1 of 1 approvals obtained

PR #810, opened 2021-04-09
Adds two new "mlab-ns" alerts for Neubot

nkinkade:
There is [an open PR in the mlab-ns repo](https://github.com/m-lab/mlab-ns/pull/251) which will cause mlab-ns to ignore monitoring data (i.e., leave datastore stale) when too large a percent of a service appears to be offline. It is important that an alert fires whenever this occurs, else mlab-ns will be serving stale data without anyone knowing. We already have alerts for various NDT services, but are currently lacking Neubot alerts. This PR adds alerts for Neubot to avoid the case where mlab-ns ignores monitoring data, and we get no alert.

This PR also updates the `PrometheusQueries` dashboard, adding panels for Neubot.

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/prometheus-support/810)

### DIFF
--- a/config/federation/grafana/dashboards/MLABNS_PrometheusQueries.json
+++ b/config/federation/grafana/dashboards/MLABNS_PrometheusQueries.json
@@ -16,7 +16,7 @@
   "gnetId": null,
   "graphTooltip": 0,
   "id": 357,
-  "iteration": 1616437301517,
+  "iteration": 1617995965275,
   "links": [],
   "panels": [
     {
@@ -35,7 +35,7 @@
       "fillGradient": 0,
       "gridPos": {
         "h": 7,
-        "w": 8,
+        "w": 6,
         "x": 0,
         "y": 0
       },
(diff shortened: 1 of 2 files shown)


stephen-soltesz (approved):
:lgtm: - with a caveat.

We're now duplicating promql queries in at least three places -- dashboard, alert, and mlab-ns -- maybe others. We really want to think about a way to generate these configurations from an earlier, single authoritative source, i.e. a configuration that outputs alerts, a config for mlab-ns to read, and ... maybe something that generates the dashboard from a template? if we used JSONNET it might not be crazy to merge/output a json dashboard with extra data.

> Reviewed 2 of 2 files at r1.
> *[Reviewable](https://reviewable.io/reviews/m-lab/prometheus-support/810#-:-MY5bm-76Q-wp_UOKM_t:brla65g)* status: :shipit: complete! 1 of 1 approvals obtained

PR #817, opened 2021-04-22
Sets sessionAffinity=ClientIP for script-exporter service

nkinkade:
The impetus for this change is how our e2e test caching works. The caching is machine-local, but the script-exporter Deployment has two replicas. You never know which replica is going to get the request from Prometheus. This means that an e2e test for a given machine is going to happen at least twice as often as we expect, and not just roughly once every 5 minutes or so. This change should cause all tests, or at least a majority, to go to one machine, allowing the caching to work more as intended. The default session affinity timeout is 10800 (3h), which means that the bulk of tests should occasionally flip flop back and forth between the replicas in the Deployment.

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/prometheus-support/817)

### DIFF
--- a/k8s/prometheus-federation/services/script-exporter.yml
+++ b/k8s/prometheus-federation/services/script-exporter.yml
@@ -12,5 +12,5 @@ spec:
     targetPort: 9172
   selector:
     run: script-exporter
-  sessionAffinity: None
+  sessionAffinity: ClientIP
   type: ClusterIP


stephen-soltesz (approved):
> *[Reviewable](https://reviewable.io/reviews/m-lab/prometheus-support/817)* status: :shipit: complete! 1 of 1 approvals obtained

PR #819, opened 2021-05-19
Add uplink speeds variable, so we can look at either 10g or 1g

gfr10598:
This change is [Reviewable](https://reviewable.io/reviews/m-lab/prometheus-support/819)

### DIFF
--- a/config/federation/grafana/dashboards/Ops_SwitchOverview.json
+++ b/config/federation/grafana/dashboards/Ops_SwitchOverview.json
@@ -837,33 +837,25 @@
           "text": "All",
           "value": "$__all"
         },
+        "datasource": "$datasource",
+        "definition": "label_values(speed)",
         "error": null,
         "hide": 0,
         "includeAll": true,
         "label": "uplink speed",
         "multi": false,
         "name": "speed",
-        "options": [
-          {
-            "selected": true,
-            "text": "All",
-            "value": "$__all"
-          },
-          {
-            "selected": false,
-            "text": "1g",
-            "value": "1g"
-          },
-          {
-            "selected": false,
-            "text": "10g",
-            "value": "10g"
-          }
-        ],
-        "query": "1g,10g",
-        "queryValue": "",
+        "options": [],
+        "query": "label_values(speed)",
+        "refresh": 2,
+        "regex": "",
         "skipUrlSync": false,
-        "type": "custom"
+        "sort": 0,
+        "tagValuesQuery": "",
+        "tags": [],
+        "tagsQuery": "",
+        "type": "query",
+        "useTags": false
       }
     ]
   },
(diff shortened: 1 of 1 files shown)


nkinkade (commented):
Comments:
* You are using a regexp matching operator for the speed label (`speed=~"$Uplinks"`). As of now, the values will always be exactly "1g" or "10g". I don't think you need the regexp operator.
* On the "Uplink dropped packets" panel, you changed it to take `max(site,speed,interface)` (previously was just `max(site)`). I think the original is still the right one. `interface` will always be the same for 10g sites (xe-0/0/45), and for 1g sites (ge-0/0/47). Selecting on `speed` should be enough. If you find it helpful, you can still leave those values in the legend, but think it is more or less a no-op in the `max()` function.
* Adding on the latter part of the previous comment, to me, adding the speed to the legends makes them a bit more clunky looking. Maybe consider adding the speed to the title of each panel instead? 

> *[Reviewable](https://reviewable.io/reviews/m-lab/prometheus-support/819#-:-Ma5WPkW5jij2Jw5mocW:b-jsv230)* status: 0 of 1 approvals obtained

gfr10598:
Ah - I misunderstood what you were recommending re: the regexp.  Changed it back, but removed the regexp.

I've made a couple other tweaks inspired by the $speed changes.  Could you PTAL?  Should show up in sandbox in about 10 minutes.

nkinkade (commented):
@gfr10598: thank you for these updates. Sorry to drag this out over small details. A few more comments:

* Thank you for updating the "speed" variable to use a query instead of static values.
* Very tiny nit: would you capitalize "uplink" in the new variable label so that it reads "Uplink speed"?
* It looks like you have added a new "Global average" metric to the top right panel. To me, this is an awkward location for this metric, since that panel is supposed to represent metrics only from the selected site. Would it make more sense to add a new panel for averages, if that is something you would like to see?
* In the "Top 10 Mbps Out" panel you are doing `max by (site, speed)`, but in the "Top 10 Mbps In" you are doing `max by (site)`. They should at least be consistent, and I think that `max by (site)` is sufficient.
* In the "Top 10 Mbps Out" panel the Legend is `{{site}} {{speed}}`, but in the "Top 10 Mbps In" the Legend is only`{{site}}`. They should probably be consistent, and maybe put `{{speed}}` in parens to differentiate the value a bit more, like `{{site}} ({{speed}})`?

> *[Reviewable](https://reviewable.io/reviews/m-lab/prometheus-support/819#-:-MaUJSd5CR5JH94FltJ9:brctgte)* status: :shipit: complete! 1 of 1 approvals obtained

nkinkade (approved):
:lgtm:

> Reviewed 1 of 1 files at r2.
> *[Reviewable](https://reviewable.io/reviews/m-lab/prometheus-support/819#-:-MaV7pfQ6C8VStysEHNN:bnfp4nl)* status: :shipit: complete! 1 of 1 approvals obtained

PR #820, opened 2021-05-20
Sandbox more panels

gfr10598:
This adds two more panels for the legacy pipeline.
Test Parsing Error Count by Type/Status: previously in Pipeline Batch (PR) and other legacy dashboards.
Insertion Error Rate and Latency:  adapted from similar panel in Pipeline Batch (PR)

Testing: - manual evaluation after pushing to sandbox.

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/prometheus-support/820)


nkinkade:
@stephen-soltesz: Seems that this PR can probably be safely closed. If we've gone without these changes for the past 2 years, then we can probably continue to do so. And fixing conflicts with the main branch could be difficult.

PR #829, opened 2021-06-25
Fixes the 'cluster' label value for platform scrape job

nkinkade:
With the move of platform alerts to the platform cluster, I also did some relabeling working. Previously, all metrics scraped from the platform clusters were given a label of `cluster="platform-cluster"`. Now that we have alerts coming from multiple clusters, alerts have a new label named `cluster` so you know where the alert originated. Presently the possible values are `prometheus-federation` and `platform`. This PR removes the redundant `-cluster` from the value of the `cluster` label so that it is more clean looking. It also fixes at least one firing alert that expects to find a label of only `cluster="platform"`.

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/prometheus-support/829)

### DIFF
--- a/config/federation/prometheus/prometheus.yml.template
+++ b/config/federation/prometheus/prometheus.yml.template
@@ -324,7 +324,7 @@ scrape_configs:
     static_configs:
       - targets: ['prometheus-platform-cluster-basicauth.{{PROJECT}}.measurementlab.net:443']
         labels:
-          cluster: "platform-cluster"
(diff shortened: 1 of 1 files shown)


stephen-soltesz (approved):
> *[Reviewable](https://reviewable.io/reviews/m-lab/prometheus-support/829)* status: :shipit: complete! 1 of 1 approvals obtained

PR #835, opened 2021-08-03
Configures BBE to probe ndt7 using TLS

nkinkade:
The current blackbox-exporter module (tcp_{v4,v6}_online) will probe
port 443 only to verify that it is listening, but does do the TLS
handshake. This change will cause it to additionally do a TLS handshake,
more fully verifying the "upness" and proper configuration of the
target.

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/prometheus-support/835)

### DIFF
--- a/generate-prometheus-targets.sh
+++ b/generate-prometheus-targets.sh
@@ -48,7 +48,7 @@ for project in mlab-sandbox mlab-staging mlab-oti ; do
       --sites "${sites}" \
       --template_target={{hostname}}:443 \
       --label service=ndt7 \
-      --label module=tcp_v4_online \
+      --label module=tcp_v4_tls_online \
       --project "${project}" \
       --select "ndt.iupui" > \
           ${output}/blackbox-targets/ndt7.json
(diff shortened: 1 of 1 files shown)


stephen-soltesz (approved):
Just to double check -- does anything depend on the old metric names/labels? Will mlab-ns queries continue to WAI with this new set of metrics? (or do we need to add both, transition, and remove the old?)

if all will WAI, then :lgtm:

> *[Reviewable](https://reviewable.io/reviews/m-lab/prometheus-support/835#-:-MgCWS7R9bK4Y3HlYAZx:b-ss9own)* status: :shipit: complete! 1 of 1 approvals obtained

nkinkade (commented):
These are very good question... thank you for considering them. I just did a quick check, and mlab-ns only relies on the `service` label, not the `module` label for all queries (most importantly the ndt7 ones):

https://github.com/m-lab/prometheus-support/pull/835

... and a recursive grep for the string `tcp_v4_online` in the entire prometheus-support repo does not turn up any results that look like an issue:

```
git/prometheus-support$ grep -r tcp_v4_online .
./generate-prometheus-targets.sh:      --label module=tcp_v4_online \
./generate-prometheus-targets.sh:      --label module=tcp_v4_online \
./config/local/blackbox/config.yml:  tcp_v4_online:
./config/federation/blackbox/config.yml.template:  tcp_v4_online:
```

I _think_ we are safe.

> *[Reviewable](https://reviewable.io/reviews/m-lab/prometheus-support/835#-:-MgCa_UUCJ1LpfK4uhkx:b-ex3r42)* status: :shipit: complete! 1 of 1 approvals obtained

PR #838, opened 2021-08-11
Use unless on() for all conditions of ipv6 alert query

stephen-soltesz:
This change updates the `PlatformCluster_IPv6AtSiteDownForTooLong` alert query to use the `unless on(machine)` operator for all conditions.

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/prometheus-support/838)

### DIFF
--- a/config/federation/prometheus/alerts.yml
+++ b/config/federation/prometheus/alerts.yml
@@ -493,10 +493,8 @@ groups:
       count by (site) (
         probe_success{service="ndt7_ipv6", ipv6="present"} == 0
           unless on(machine) probe_success{service="ndt7"} == 0
-          unless on(machine) (
-            gmx_machine_maintenance == 1 or
-            kube_node_spec_taint{cluster="platform", key="lame-duck"}
-          )
+          unless on(machine) gmx_machine_maintenance == 1
+          unless on(machine) kube_node_spec_taint{cluster="platform", key="lame-duck"}
       ) > 0
     for: 24h
     labels:
--- a/config/federation/prometheus/alerts.yml
+++ b/config/federation/prometheus/alerts.yml
@@ -492,9 +492,11 @@ groups:
     expr: |
       count by (site) (
         probe_success{service="ndt7_ipv6", ipv6="present"} == 0
-          unless on(machine) probe_success{service="ndt7"} == 0
-          unless on(machine) gmx_machine_maintenance == 1
-          unless on(machine) kube_node_spec_taint{cluster="platform", key="lame-duck"}
+          unless on(machine) (
+            probe_success{service="ndt7"} == 0
+            or gmx_machine_maintenance == 1
+            or kube_node_spec_taint{cluster="platform", key="lame-duck"}
+          )
       ) > 0
     for: 24h
     labels:


stephen-soltesz (commented):
> *[Reviewable](https://reviewable.io/reviews/m-lab/prometheus-support/838)* status: 1 change requests, 0 of 1 approvals obtained (waiting on @nkinkade)

___
*[config/federation/prometheus/alerts.yml, line 497 at r1](https://reviewable.io/reviews/m-lab/prometheus-support/838#-Mgpx2so9kurqJui5b5S:-Mgq1SfBCBBFAcOVJc0r:b-bj16h9) ([raw file](https://github.com/m-lab/prometheus-support/blob/079e40f5c159fd71daa601867b7f2019b38f918f/config/federation/prometheus/alerts.yml#L497)):*

Our of curiosity, would it not be more succinct as:

```
count by (site) (
  probe_success{service="ndt7_ipv6", ipv6="present"} == 0
    unless on(machine) (
      probe_success{service="ndt7"} == 0 or
      gmx_machine_maintenance == 1 or
      kube_node_spec_taint{cluster="platform", key="lame-duck"}
    )
 )  

</blockquote></details>

Your question makes sense. I have found the semantics of `OR` expressions confusing in the past. OR expressions only return one timeseries in the expression under some conditions (the worst case was the [daily parser alert rule][1]). So, my thinking in this case is that it's easier to reason about the `unless on()` chain than it is to reason about and be sure that `or` expressions are correct.

[1]: https://github.com/m-lab/prometheus-support/blob/master/config/federation/prometheus/alerts.yml#L667


<!-- Sent from Reviewable.io -->

nkinkade (approved):
> *[Reviewable](https://reviewable.io/reviews/m-lab/prometheus-support/838)* status: :shipit: complete! 1 of 1 approvals obtained (waiting on @stephen-soltesz)

___
*[config/federation/prometheus/alerts.yml, line 497 at r1](https://reviewable.io/reviews/m-lab/prometheus-support/838#-Mgpx2so9kurqJui5b5S:-Mgq3kRS1nzbeTb1LEyY:bjow2e5) ([raw file](https://github.com/m-lab/prometheus-support/blob/079e40f5c159fd71daa601867b7f2019b38f918f/config/federation/prometheus/alerts.yml#L497)):*

Your question makes sense. I have found the semantics of `OR` expressions confusing in the past. OR expressions only return one timeseries in the expression under some conditions (the worst case was the [daily parser alert rule][1]). So, my thinking in this case is that it's easier to reason about the `unless on()` chain than it is to reason about and be sure that `or` expressions are correct.

[1]: https://github.com/m-lab/prometheus-support/blob/master/config/federation/prometheus/alerts.yml#L667
</blockquote></details>

My understanding is that the implicit AND operator returns an intersection of the expressions, while an explicit OR returns a union. In my proposed change, if a metric does not exist it will not factor into the `unless on()`, and I believe the same is true for your version. I believe the two versions are functionally equivalent. I have a slight preference for my suggestion (of course), but I don't have a strong opinion about it either way, so :lgtm:.

stephen-soltesz (commented):
> *[Reviewable](https://reviewable.io/reviews/m-lab/prometheus-support/838)* status: :shipit: complete! 1 of 1 approvals obtained (waiting on @nkinkade)

___
*[config/federation/prometheus/alerts.yml, line 497 at r1](https://reviewable.io/reviews/m-lab/prometheus-support/838#-Mgpx2so9kurqJui5b5S:-Mgqhwnq75umJbRIV1Ol:b-y9baex) ([raw file](https://github.com/m-lab/prometheus-support/blob/079e40f5c159fd71daa601867b7f2019b38f918f/config/federation/prometheus/alerts.yml#L497)):*

My understanding is that the implicit AND operator returns an intersection of the expressions, while an explicit OR returns a union. In my proposed change, if a metric does not exist it will not factor into the `unless on()`, and I believe the same is true for your version. I believe the two versions are functionally equivalent. I have a slight preference for my suggestion (of course), but I don't have a strong opinion about it either way, so :lgtm:. 
</blockquote></details>

* I want all the toppings, unless they're red, or unless they're blue, or unless they're green.
* I want all the toppings, unless they're (red, or blue, or green).

Okay -- it does seem simpler using the `or` construct. And the scenario I'm worried about where `or` is confusing should not occur here.

PR #868, opened 2022-01-26
Upgrades Grafana to v8.3.4

nkinkade:
Moving to Grafana v8.3.4 required one particular change in our datasource definitions that use HTTP basic auth.  The configuration `basicAuthPassword` was deprecated and was replaced by `secureJsonData.basicAuthPassword`.

Additionally, while troubleshooting the issues caused by the above, I found that our mostly default Grafana configuration file was too long, outdated, and mostly just commented out defaults. To make it much easier to read, I removed every single commented out item, leaving only the options which we modify that are different from defaults. This reduced the document to just a dozen or so line, down from hundreds. At the top of this now-abbreviated config file is a link to a sample config file from upstream with all the possible configurations and options.

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/prometheus-support/868)


robertodauria (approved):
:lgtm: - nice!

> Reviewed 8 of 8 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/m-lab/prometheus-support/868#-:-MuQYK6y6Ps_cBzvucjH:bgw2fs2)* status: :shipit: complete! 1 of 1 approvals obtained

PR #869, opened 2022-01-26
Update or prepare bigquery exporter queries to use v2 ndt5 tables

stephen-soltesz:
This change updates or prepares existing bigquery exporter queries for the pending changes to the ndt5 tables.

Since production will be updated last, the references to `measurement-lab.ndt.ndt5` are replaced with references to (the equivalent) `measurement-lab.ndt_raw.ndt5_legacy`. This will allow the production deployment of the new table without breaking these queries before deployment. However, after production deployment, these references to `ndt5_legacy` will be come stale, as the v1 parser and gardener will no longer be populating the base_tables.ndt5 v1 tables. For the `bq_daily_tests.sql ` query, this will result in an underestimate of the ndt5 rows until the update is deployed.

Part of:

* https://github.com/m-lab/etl/issues/1046

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/prometheus-support/869)


cristinaleonr (approved):
I assume [this alert] (https://github.com/m-lab/prometheus-support/blob/02e6fcfaa32db1c039fd68110f242551685c072e/config/federation/prometheus/alerts.yml#L7441) that uses `bq_daily_tests` won't misfire while the update is deployed because it has to do with switch discards and not the actual number of ndt5 rows.

:lgtm: 

> *[Reviewable](https://reviewable.io/reviews/m-lab/prometheus-support/869#-:-MuOYgcIEaSQmMQe5uOM:bg5ryo5)* status: :shipit: complete! 1 of 1 approvals obtained

PR #875, opened 2022-02-08
Sets correct datasource for panel

nkinkade:
The "Excessive 6h/90th Percentile of Switch Capcity" panel was using the
prometheus-federation datasource, but quite some time ago the required
metric were moved to the platform cluster. This commit sets the correct
datasource for the panel.

This small change to the dashboard has brought in a lot more changes than are apparently needed to just fix the one issue which it aims to correct. This must have to do with the JSON being generated by a newer version of Grafana than it was originally created with.

### DIFF
--- a/config/federation/grafana/dashboards/Ops_Tactical_SRE_Overview.json
+++ b/config/federation/grafana/dashboards/Ops_Tactical_SRE_Overview.json
@@ -14,21 +14,23 @@
   },
   "description": "This dashboard displays metrics that are useful to be reviewed as part of the Ops Tactical meetings and SRE rotation.",
   "editable": true,
-  "gnetId": null,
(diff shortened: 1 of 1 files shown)


stephen-soltesz (approved):
I verified the metric appears in the mlab-sandbox SRE overview dashboard for the switch panel (when I selected the correct datasource and removed the thresholds -- so the metric is there and presumably if the threshold was surpassed it would appear normally).

Thank you for the changes!

:lgtm:

> *[Reviewable](https://reviewable.io/reviews/m-lab/prometheus-support/875#-:-MvQ0ojb6mwRFXIzcb6t:bxqszxn)* status: :shipit: complete! 1 of 1 approvals obtained

PR #881, opened 2022-02-25
Adds new alert to monitor Vector functionality

nkinkade:
This PR adds a new alert which aims to give us some confidence that Vector in the platform clusters is working as intended. A new logs-based metric named platform-cluster-container-logs was created, and we query this new metric to determine if Vector is pushing container logs to Stackdriver. The new metric is a count of log entries on a per-node basis. When the sum of that count over an hour is equal to 0 for 24h, then this alert should fire.

This PR also updates the mlab-ns instance of stackdriver_exporter to v0.11.0, and moves to using the proper Docker repository.

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/prometheus-support/881)

### DIFF
--- a/config/federation/prometheus/alerts.yml
+++ b/config/federation/prometheus/alerts.yml
@@ -548,6 +548,24 @@ groups:
         the underlying service. Please check the GAE dashboard and logs to
         determine the cause.
 
+# If container logs for a node are missing in Stackdriver for too long, then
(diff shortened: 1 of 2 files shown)


stephen-soltesz (approved):
> *[Reviewable](https://reviewable.io/reviews/m-lab/prometheus-support/881)* status: :shipit: complete! 1 of 1 approvals obtained

PR #883, opened 2022-03-01
Updates alert for when Stackdriver is missing container logs for a node

nkinkade:
The current alert is based on a misunderstanding I had about how metrics from Stackdriver work. This updated alert should warn us when a node's status is "Ready=True", but there is no stackdriver metric indicating that any log entries were received by Stackdriver for a node.

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/prometheus-support/883)

### DIFF
--- a/config/federation/prometheus/alerts.yml
+++ b/config/federation/prometheus/alerts.yml
@@ -559,7 +559,7 @@ groups:
         unless on(node) label_replace(
           sum_over_time(
             stackdriver_generic_node_logging_googleapis_com_user_platform_cluster_container_logs
-          [15m]),
+          [1h]),
           "node", "$1", "node_id", "(.*)"
         )
         unless on(node) gmx_machine_maintenance == 1
--- a/config/federation/prometheus/alerts.yml
+++ b/config/federation/prometheus/alerts.yml
@@ -556,13 +556,10 @@ groups:
   - alert: PlatformCluster_ContainerLogsMissingInStackdriver
     expr: |
       kube_node_status_condition{condition="Ready", status="true", node=~"mlab[1-4].*"} == 1
-        unless on(node) label_replace(
-          sum_over_time(
-            stackdriver_generic_node_logging_googleapis_com_user_platform_cluster_container_logs
-          [1h]),
-          "node", "$1", "node_id", "(.*)"
-        )
-        unless on(node) gmx_machine_maintenance == 1
+        unless on(machine) sum_over_time(
+          stackdriver_generic_node_logging_googleapis_com_user_platform_cluster_container_logs
+        [1h])
+        unless on(machine) gmx_machine_maintenance == 1
     for: 24h
     labels:
       repo: ops-tracker
--- a/config/federation/prometheus/alerts.yml
+++ b/config/federation/prometheus/alerts.yml
@@ -549,13 +549,20 @@ groups:
         determine the cause.
 
 # If container logs for a node are missing in Stackdriver for too long, then
-# fire an alert, unless the node is in maintenance mode.
+# fire an alert, unless the node is in maintenance mode. This somewhat awkward
+# query discovers cases where the stackdriver metric has been absent for too
+# long, indicating a likely issue with the Vector pod on the node, causing it
+# to fail to upload container logs to Stackdriver.
   - alert: PlatformCluster_ContainerLogsMissingInStackdriver
     expr: |
-      label_replace(
-        sum_over_time(stackdriver_generic_node_logging_googleapis_com_user_platform_cluster_container_logs[1h]) == 0,
-        "node", "$1", "node_id", "(.*)"
-      ) unless on(node) gmx_machine_maintenance == 1
+      kube_node_status_condition{condition="Ready", status="true", node=~"mlab[1-4].*"} == 1
+        unless on(node) label_replace(
+          sum_over_time(
+            stackdriver_generic_node_logging_googleapis_com_user_platform_cluster_container_logs
+          [15m]),
+          "node", "$1", "node_id", "(.*)"
+        )
+        unless on(node) gmx_machine_maintenance == 1
     for: 24h
     labels:
       repo: ops-tracker
--- a/k8s/prometheus-federation/deployments/github-receiver.yml
+++ b/k8s/prometheus-federation/deployments/github-receiver.yml
@@ -17,7 +17,7 @@ spec:
     spec:
       containers:
       - name: github-receiver
-        image: measurementlab/alertmanager-github-receiver:v0.8
+        image: measurementlab/alertmanager-github-receiver:v0.10
         env:
         - name: GITHUB_AUTH_TOKEN
           valueFrom:


stephen-soltesz (approved):
Sorry, I think I missed this one.

:lgtm: - with a possible improvement.

> Reviewed 1 of 2 files at r1.
> *[Reviewable](https://reviewable.io/reviews/m-lab/prometheus-support/883#-:-MxGOVXj2CQ0Fx85fwMp:b-fo0ml4)* status: :shipit: complete! 1 of 1 approvals obtained (waiting on @nkinkade)

___
*[config/federation/prometheus/alerts.yml, line 563 at r1](https://reviewable.io/reviews/m-lab/prometheus-support/883#-MxGOjkr7TB511yWHQTj:-MxGOjkr7TB511yWHQTk:b-hb1def) ([raw file](https://github.com/m-lab/prometheus-support/blob/1086d0cf02ac4190c94c9ff390ca73555c095708/config/federation/prometheus/alerts.yml#L563)):*
> ```YAML
>             stackdriver_generic_node_logging_googleapis_com_user_platform_cluster_container_logs
>           [1h]),
>           "node", "$1", "node_id", "(.*)"
> ```

Is there any chance the label assignment of "node" can be done in the stackdriver config? (this is the custom metric definition, right?) The idea being reducing the complexity of the promql query if possible.

nkinkade (commented):
> *[Reviewable](https://reviewable.io/reviews/m-lab/prometheus-support/883)* status: :shipit: complete! 1 of 1 approvals obtained (waiting on @stephen-soltesz)

___
*[config/federation/prometheus/alerts.yml, line 563 at r1](https://reviewable.io/reviews/m-lab/prometheus-support/883#-MxGOjkr7TB511yWHQTj:-MxGfoGZ8kP5acYNT6Wx:buo79hc) ([raw file](https://github.com/m-lab/prometheus-support/blob/1086d0cf02ac4190c94c9ff390ca73555c095708/config/federation/prometheus/alerts.yml#L563)):*

Is there any chance the label assignment of "node" can be done in the stackdriver config? (this is the custom metric definition, right?) The idea being reducing the complexity of the promql query if possible.
</blockquote></details>

@stephen-soltesz: My first iteration of the log-based metric did include a custom`node` label extracted from the field `node_id`. But for some reason, the stackdriver exporter was crashlooping because of, I believe, this issue:

https://github.com/prometheus-community/stackdriver_exporter/issues/85

I deleted the label and it started running normally. I don't know why that label was causing the error, since, as far as I can tell, there are no labels named "node" added by GCP.

Another possible solution would be to have Vector push the log with a label of "node". For reasons I don't know, it is currently pushing the label "node_id":

https://github.com/m-lab/k8s-support/blob/master/config/vector/values.yaml.template#L17

I _believe_ those labels are arbitrary and that we could simply renamed "node_id" to "node", problem solved. Do you happen to know if those labels are predefined somehow in GCP logging?

nkinkade (commented):
> *[Reviewable](https://reviewable.io/reviews/m-lab/prometheus-support/883)* status: :shipit: complete! 1 of 1 approvals obtained (waiting on @stephen-soltesz)

___
*[config/federation/prometheus/alerts.yml, line 563 at r1](https://reviewable.io/reviews/m-lab/prometheus-support/883#-MxGOjkr7TB511yWHQTj:-MxGifHd7lbm5Ux6pJIl:b4gbs8t) ([raw file](https://github.com/m-lab/prometheus-support/blob/1086d0cf02ac4190c94c9ff390ca73555c095708/config/federation/prometheus/alerts.yml#L563)):*

@stephen-soltesz: My first iteration of the log-based metric did include a custom`node` label extracted from the field `node_id`. But for some reason, the stackdriver exporter was crashlooping because of, I believe, this issue:

https://github.com/prometheus-community/stackdriver_exporter/issues/85

I deleted the label and it started running normally. I don't know why that label was causing the error, since, as far as I can tell, there are no labels named "node" added by GCP.

Another possible solution would be to have Vector push the log with a label of "node". For reasons I don't know, it is currently pushing the label "node_id":

https://github.com/m-lab/k8s-support/blob/master/config/vector/values.yaml.template#L17

I _believe_ those labels are arbitrary and that we could simply renamed "node_id" to "node", problem solved. Do you happen to know if those labels are predefined somehow in GCP logging?
</blockquote></details>

Oh, I see where the duplicate `node` label is/was coming from. It seems that Stackdriver _is_ already adding a "node" label, but with values like:

```
node="gke-prometheus-feder-prometheus-pool--2df54f21-lns5"
```

... presumably the node in GKE where stackdriver_exporter exporter is running?

Let me play around with having Vector push a "node" label instead of "node_id" to see what GCP does with that, since it may conflict with a label it wasn't to add itself. The value that Vector pushes may get overwritten by GCP.

nkinkade (commented):
> *[Reviewable](https://reviewable.io/reviews/m-lab/prometheus-support/883)* status: :shipit: complete! 1 of 1 approvals obtained (waiting on @stephen-soltesz)

___
*[config/federation/prometheus/alerts.yml, line 563 at r1](https://reviewable.io/reviews/m-lab/prometheus-support/883#-MxGOjkr7TB511yWHQTj:-MxGrMkECGUqsncUSUtQ:b-hd6qqq) ([raw file](https://github.com/m-lab/prometheus-support/blob/1086d0cf02ac4190c94c9ff390ca73555c095708/config/federation/prometheus/alerts.yml#L563)):*

Oh, I see where the duplicate `node` label is/was coming from. It seems that Stackdriver _is_ already adding a "node" label, but with values like:

```
node="gke-prometheus-feder-prometheus-pool--2df54f21-lns5"
```

... presumably the node in GKE where stackdriver_exporter exporter is running?

Let me play around with having Vector push a "node" label instead of "node_id" to see what GCP does with that, since it may conflict with a label it wasn't to add itself. The value that Vector pushes may get overwritten by GCP.
</blockquote></details>

It would appear that the resource "generic_node" only accepts certain labels:

https://cloud.google.com/monitoring/api/resources#tag_generic_node

Interestingly, I once again added a custom "node" label to the log-based metric, and this time stackdriver_exporter has not crashed. However, nor does the "node" label come through with the custom label. The node label still has the GKE node name, not the M-Lab node name.

nkinkade (commented):
> *[Reviewable](https://reviewable.io/reviews/m-lab/prometheus-support/883)* status: :shipit: complete! 1 of 1 approvals obtained (waiting on @stephen-soltesz)

___
*[config/federation/prometheus/alerts.yml, line 563 at r1](https://reviewable.io/reviews/m-lab/prometheus-support/883#-MxGOjkr7TB511yWHQTj:-MxKoBf4ClT8rSFHmLN0:b-5ndzi) ([raw file](https://github.com/m-lab/prometheus-support/blob/1086d0cf02ac4190c94c9ff390ca73555c095708/config/federation/prometheus/alerts.yml#L563)):*

It would appear that the resource "generic_node" only accepts certain labels:

https://cloud.google.com/monitoring/api/resources#tag_generic_node

Interestingly, I once again added a custom "node" label to the log-based metric, and this time stackdriver_exporter has not crashed. However, nor does the "node" label come through with the custom label. The node label still has the GKE node name, not the M-Lab node name.
</blockquote></details>

@stephen-soltesz: I was unable to modify the `node` label in the metric. It appears that GCP adds that label, and it can't be overriden. I was, however, able to add a `machine` label to the metric. Most all of our metrics have the `machine` label. This works, and makes the query a bit simpler. In my mind the slight downside is that we are now using the `machine` label, which, in my mind, is sort of a legacy label name that I've been want to replace with the `node` label.

What do you think?

stephen-soltesz (approved):
> *[Reviewable](https://reviewable.io/reviews/m-lab/prometheus-support/883)* status: :shipit: complete! 1 of 1 approvals obtained

___
*[config/federation/prometheus/alerts.yml, line 563 at r1](https://reviewable.io/reviews/m-lab/prometheus-support/883#-MxGOjkr7TB511yWHQTj:-Mx_hsMkEp1ck2LqHafa:b-nppkdy) ([raw file](https://github.com/m-lab/prometheus-support/blob/1086d0cf02ac4190c94c9ff390ca73555c095708/config/federation/prometheus/alerts.yml#L563)):*

@stephen-soltesz: I was unable to modify the `node` label in the metric. It appears that GCP adds that label, and it can't be overriden. I was, however, able to add a `machine` label to the metric. Most all of our metrics have the `machine` label. This works, and makes the query a bit simpler. In my mind the slight downside is that we are now using the `machine` label, which, in my mind, is sort of a legacy label name that I've been want to replace with the `node` label.

What do you think?
</blockquote></details>

You eliminated the `label_replace` which is a win in my book.

It looks like there is a general need to differentiate between machine/node as it relates to the fleet we manage and the pieces of the system that are managed by GKE or Kubernetes generally. While "machine" may be vestigial, keeping it for a reason like this may be advantageous, b/c it provides away to avoid collisions with "node" across our cluster, gke clusters, and parts of GCP. If this is a general approach that we want to adopt more widely, we should document this convention.

PR #880, opened 2022-02-23
Update bq exporter queries to use the v2 switch datatype

robertodauria:
This PR updates the bigquery exporter queries to refer to the v2 switch tables and to use the new schema.

Note: I won't merge this until 1. the v2 switch datatype parser is merged and 2. the views referenced in this PR have been created (through a separate PR to etl-schema).

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/prometheus-support/880)

PR #885, opened 2022-03-08
Update bq_ndt_s2c.sql to use v2 datatype schema

robertodauria:
This PR reintroduces the query change that was excluded from https://github.com/m-lab/prometheus-support/pull/880 so that we could release prometheus-support. This change should not be merged until etl-schema is released, since the `measurement-lab.utilization.switch` view won't exist until then.

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/prometheus-support/885)


robertodauria:
- etl-schema has been released a few days ago (https://github.com/m-lab/etl-schema/releases/tag/v3.36.0).
- I've updated the two references to `ndt_raw.ndt5_legacy` to to `ndt.ndt5`

PR #886, opened 2022-03-08
Updates to SiteOverview to better represent virtual sites

nkinkade:
The diff is pretty large for this PR, apparently mostly containing changes that Grafana automatically added in the exported JSON due to this dashboard having been created with some older version of Grafana.

There are only two changes I made:

* Update the host network panel to look for data for both `eth0` _and_ `ens4` (virtual nodes).
* Updates the "Data partition" singlestat panel to display the usage of the root device (/dev/root) for virtual nodes. This will yield the same value as the "Rootfs" panel, but may(?) reduce confusion, since virtual nodes do everything on a single disk. That is, the rootfs and the "data partition" are one and the same.

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/prometheus-support/886)

### DIFF
--- a/config/federation/grafana/dashboards/K8s_SiteOverview.json
+++ b/config/federation/grafana/dashboards/K8s_SiteOverview.json
@@ -8,20 +8,26 @@
         "hide": true,
         "iconColor": "rgba(0, 211, 255, 1)",
         "name": "Annotations & Alerts",
+        "target": {
+          "limit": 100,
+          "matchAny": false,
+          "tags": [],
+          "type": "dashboard"
+        },
         "type": "dashboard"
       }
     ]
   },
   "editable": true,
-  "gnetId": null,
+  "fiscalYearStartMonth": 0,
   "graphTooltip": 1,
   "id": 246,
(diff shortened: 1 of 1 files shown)


nkinkade (commented):
@stephen-soltesz: I added one more commit to add a "Machine type" stat panel. Currently, it only works in sandbox, until I submit another small PR to you for k8s-support. But you can check that it works in sandbox for now.

> *[Reviewable](https://reviewable.io/reviews/m-lab/prometheus-support/886#-:-MxfA6QLBSHiP-_i1eW6:bifraq0)* status: 0 of 1 approvals obtained (waiting on @stephen-soltesz)

stephen-soltesz (approved):
:lgtm: - we discussed on VC that the site drop down does not currently include cloud sites in sandbox - reason unknown.

> *[Reviewable](https://reviewable.io/reviews/m-lab/prometheus-support/886#-:-MxfErq-Dh4bh-cwrB3U:b-iiu3pf)* status: :shipit: complete! 1 of 1 approvals obtained

PR #898, opened 2022-04-23
Allows selecting on site type in NDT:GlobalTestRate

nkinkade:
The user can select either "Physical" or "Virtual", or both. The default is both selected.

### DIFF
--- a/config/federation/grafana/dashboards/NDT_GlobalTestRate.json
+++ b/config/federation/grafana/dashboards/NDT_GlobalTestRate.json
@@ -13,14 +13,14 @@
     ]
   },
   "editable": true,
-  "gnetId": null,
+  "fiscalYearStartMonth": 0,
   "graphTooltip": 1,
-  "iteration": 1592444440456,
+  "id": 195,
+  "iteration": 1650662213697,
   "links": [],
+  "liveNow": false,
   "panels": [
     {
-      "content": "# Real time",
-      "datasource": null,
       "gridPos": {
         "h": 3,
         "w": 24,
@@ -29,9 +29,11 @@
       },
       "id": 21,
       "links": [],
-      "mode": "markdown",
-      "options": {},
-      "title": "",
+      "options": {
+        "content": "# Real time",
+        "mode": "markdown"
+      },
+      "pluginVersion": "8.3.4",
       "type": "text"
     },
     {
@@ -39,8 +41,17 @@
       "bars": false,
       "dashLength": 10,
       "dashes": false,
-      "datasource": "$datasource",
+      "datasource": {
+        "type": "prometheus",
+        "uid": "$datasource"
+      },
       "decimals": 1,
+      "fieldConfig": {
+        "defaults": {
+          "links": []
+        },
+        "overrides": []
+      },
       "fill": 0,
       "fillGradient": 0,
       "gridPos": {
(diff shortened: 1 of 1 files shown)


stephen-soltesz (commented):
Waiting for feedback on the "workload" filter.

> *[Reviewable](https://reviewable.io/reviews/m-lab/prometheus-support/898#-:-N0VfhKnDg6qSWHlwxSB:b-snfxzx)* status: 0 of 1 approvals obtained

nkinkade:
It turns out that using the `workload` label for differentiating metrics from physical and virtual sites will not work (at least easily). The problem is that our NDT canary workloads are named like `ndt-<desc>` (e.g., "ndt-osupgrade", "ndt-canary", etc.). Using only PromQL it seems like it may not be possible to have a simple select box with only two options ("Physical", "Virtual") which is able to be smart enough to select all workloads like "ndt", "ndt-osupgrade", "ndt-canary", while ignoring pods with the label "ndt-virtual".

I have submitted a new PR to the k8s-support repo which adds explicit pod labels for what type of site the pods is running on. When/if that PR is merged, then I can update this PR to start selecting on this new pod label (`measurementlab.net/type`) rather than on `workload`.

nkinkade:
The PR to k8s-support was merged. I have now updated this PR to select on the new `measurementlab.net/type` metric label.

@stephen-soltesz: PTAL, after the Travis build succeeds?

nkinkade (commented):
My concern with just "type" is that, to me, it seems under specified. "Type" could mean anything. Yes, in context, a user would say: okay, `type="virtual"`, then "type" here must mean machine type... but it's not impossible that we may want to specify some other attribute in the future that could also be thought of a "type" of something.

Out of curiosity, do you also think the `mlab/type` node label is not ideal, which has the same meaning? We could just reuse that same label for pods.

`platform_type` doesn't seem quite right either. When I think of "platform", I am thinking of something like bare-metal, GCP, Azure, AWS, etc... not just the basic distinction between virtual and physical. A label like that (maybe just plain `platform` instead) could definitely be useful in the future.

The k8s documentation on labels used the word "must" when saying that the optional prefix be a domain name. It, of course, doesn't mean we have to do that, but when I read the word "must" in official documentation, it feels off to ignore it and go our own route and to interpret it loosely. It makes me consider/wonder if at some point in the future k8s could enforce that rule.

Since we are talking about pod labels, maybe we could use a label of `machine_type`, to convey what sort of machine the workload is executing on?

> *[Reviewable](https://reviewable.io/reviews/m-lab/prometheus-support/898#-:-N0fgXfs1FA1DaAmiYfv:b-i8hs61)* status: 0 of 1 approvals obtained

stephen-soltesz (approved):
Thank you for the name change.

:lgtm:

I'm a little sad we lose the history for our global test rate panel now. But, we'll have it back soon enough.

> *[Reviewable](https://reviewable.io/reviews/m-lab/prometheus-support/898#-:-N0hPfMq9bJFM0RFH9ce:b-tumcf2)* status: :shipit: complete! 1 of 1 approvals obtained

nkinkade (commented):
If we are very sad about it, then we could create a temporary sandbox version of the dashboard without the new label selectors, and then we'd have the old data back until the new dashboard fully populated in a few weeks, right?

> *[Reviewable](https://reviewable.io/reviews/m-lab/prometheus-support/898#-:-N0i2xuWDkjxstFg4rKk:b77wb5x)* status: :shipit: complete! 1 of 1 approvals obtained

PR #914, opened 2022-05-20
Replaces datasource UID with $datasource

nkinkade:
Somehow various panels got a statically defined datasource UID, which is
only valid in sandbox. The correct UID should be "$datasource", allowing
users to effectively switch datasources from a select box. I'm not sure
when error was introduced, but this commit should correct it.

### DIFF
--- a/config/federation/grafana/dashboards/K8s_SiteOverview.json
+++ b/config/federation/grafana/dashboards/K8s_SiteOverview.json
@@ -966,8 +966,7 @@
       "targets": [
         {
           "datasource": {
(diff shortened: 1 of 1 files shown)


stephen-soltesz (approved):
Thank you :lgtm: 

> *[Reviewable](https://reviewable.io/reviews/m-lab/prometheus-support/914#-:-N2YcfJ6-5wwyTAZgPEx:buxrqzc)* status: :shipit: complete! 1 of 1 approvals obtained

PR #916, opened 2022-05-24
Add other datatypes and set longer annotation threshold

stephen-soltesz:
After migration to the v2 data pipeline and the retirement of the annotation-service, the `DataPipeline_(Asn|Geo)AnnotationRatioTooLowOrMissing` alerts needed to be updated for v2 data type annotations. This change adds two additional annotated data types. Since uuid-annotations are collected at measurement times, and bq processing is delayed 2days, this alert should represent a failure of the pipeline rather than a data loss. Other alerts should cover incomplete uuid-annotation data volume from the node.

This change also extends the alert threshold time from 30min to 60min to allow GKE to restart pods and for the bq-exporter to rerun queries to report metrics. Previously, the 30min threshold has caused false positives due to firing prematurely.

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/prometheus-support/916)

### DIFF
--- a/config/federation/bigquery/bq_annotation.sql
+++ b/config/federation/bigquery/bq_annotation.sql
@@ -37,6 +37,10 @@ FROM (
    SELECT * FROM recent_ndt_tcpinfo
    UNION ALL
    SELECT * FROM recent_ndt_ndt7
+   UNION ALL
+   SELECT * FROM recent_ndt_ndt5
+   UNION ALL
+   SELECT * FROM recent_ndt_scamper1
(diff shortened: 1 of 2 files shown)


nkinkade (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/m-lab/prometheus-support/916#-:-N2rHm6AC29z7puCrKaB:bnfp4nl)* status: :shipit: complete! 1 of 1 approvals obtained

PR #923, opened 2022-06-17
Add average alerts last 30days

stephen-soltesz:
This change adds a new panel to the Ops Tactical SRE Overview dashboard. The new panel pins a time range for the last 30days with the average daily alert rate received by the alertmanager github receiver.

This change is part of an effort to ensure that we monitor the average alerts / day and focus on keeping this rate below 1/day.

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/prometheus-support/923)

### DIFF
--- a/config/federation/grafana/dashboards/Ops_Tactical_SRE_Overview.json
+++ b/config/federation/grafana/dashboards/Ops_Tactical_SRE_Overview.json
@@ -16,7 +16,7 @@
   "editable": true,
   "fiscalYearStartMonth": 0,
   "graphTooltip": 0,
-  "iteration": 1655494122017,
+  "iteration": 1655496781097,
   "links": [],
   "liveNow": false,
   "panels": [
@@ -751,6 +751,10 @@
       "type": "table-old"
     },
     {
+      "datasource": {
+        "type": "prometheus",
+        "uid": "${datasource}"
+      },
       "description": "",
       "fieldConfig": {
         "defaults": {
@@ -824,7 +828,7 @@
         {
           "datasource": {
             "type": "prometheus",
-            "uid": "000000008"
+            "uid": "${datasource}"
           },
           "exemplar": true,
           "expr": "sum (increase(githubreceiver_alerts_total{status=\"firing\"}[30d])) / 30",
@@ -1119,7 +1123,7 @@
     "list": [
       {
         "current": {
-          "selected": true,
+          "selected": false,
           "text": "Prometheus (mlab-oti)",
           "value": "Prometheus (mlab-oti)"
         },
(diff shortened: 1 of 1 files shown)


cristinaleonr (commented):
Hmm, I see information in all the panels except for this new one.



> *[Reviewable](https://reviewable.io/reviews/m-lab/prometheus-support/923#-:-N4nGVds1EKO2eaXEVeG:b-kfj48v)* status: 0 of 1 approvals obtained

stephen-soltesz (commented):
Awkward :) Looks like it was the datasource reference again. Fixed now. PTAL?

> *[Reviewable](https://reviewable.io/reviews/m-lab/prometheus-support/923#-:-N4nK5DGDKfpk138PQ9G:bwue6ep)* status: 0 of 1 approvals obtained

cristinaleonr (commented):
Would it make more sense for this to pick up the value from the time picker (as opposed to being hard-coded to 30 days)?

> *[Reviewable](https://reviewable.io/reviews/m-lab/prometheus-support/923#-:-N4nKeP55PRh9YMk5INH:brsal3u)* status: 0 of 1 approvals obtained

stephen-soltesz (commented):
I don't think so b/c we typically only look at the last 7 days for the SRE overview dashboard, but the average alert rate is measured at a longer timescales (30d in this case). It's the longer term trends that are currently invisible. So, I hope that taken together, the dashboard will show a near and long term window at the same time. Additional longer term trends would be welcome too imo.

> *[Reviewable](https://reviewable.io/reviews/m-lab/prometheus-support/923#-:-N4nLM5LCX8KrF_rvdKx:b-93rzf4)* status: 0 of 1 approvals obtained

cristinaleonr (approved):
My point is that it's better to have a configurable number than one that is hard-coded. That way, we can easily compare with even longer trends (e.g., 60, 90 days).

> *[Reviewable](https://reviewable.io/reviews/m-lab/prometheus-support/923#-:-N4nMx_T7igP8vk21I5Q:b-8kj7b4)* status: :shipit: complete! 1 of 1 approvals obtained

stephen-soltesz (commented):
I agree flexibility is good. This is about the power of defaults. I want the longer time periods to be surfaced/visible without any extra steps. When we review the SRE dashboard, the typical behavior is to only look at the default 7day view (which makes sense for a week to week tactical meeting). But, it the longer term trends that we're missing in that 7day view.

Is there a better way to satisfy the goals of a) seeing longer trends in a default view with b) more flexibility? I just tried adding a variable but unfortunately the query options don't allow variable substitutions. Is there another way?

> *[Reviewable](https://reviewable.io/reviews/m-lab/prometheus-support/923#-:-N4nNvsU1MPy5YDKRrpu:btjjxkw)* status: :shipit: complete! 1 of 1 approvals obtained

stephen-soltesz:
Let's give it a shot. We can always refine.

PR #932, opened 2022-08-02
Adds a new GKE Datasource variable to SiteOverview

nkinkade:
The dashboard has a few panels that make use of metrics from the
prometheus-federation cluster. Previously it was using a static
datasource UID, which was only valid for the prometheus-federation
cluster in mlab-sandbox. With this commit, we can properly select the
right datasource at the top and all panels should work as expected in
all projects.

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/prometheus-support/932)

### DIFF
--- a/config/federation/grafana/dashboards/K8s_SiteOverview.json
+++ b/config/federation/grafana/dashboards/K8s_SiteOverview.json
@@ -21,8 +21,8 @@
   "editable": true,
   "fiscalYearStartMonth": 0,
   "graphTooltip": 1,
-  "id": 65,
-  "iteration": 1659037480998,
+  "id": 246,
+  "iteration": 1659542413927,
   "links": [],
   "liveNow": false,
   "panels": [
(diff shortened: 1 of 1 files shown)


cristinaleonr (commented):
Does this datasource only use metrics from the prometheus-federation cluster? 

If so, is it possible to restrict the variable to only show prometheus-federation options (e.g., "/Prometheus \(/")?

> *[Reviewable](https://reviewable.io/reviews/m-lab/prometheus-support/932#-:-N8Vveu00T3wPo6VwU6U:b-ogpbcc)* status: 0 of 1 approvals obtained

nkinkade (commented):
Good idea. I have limited both the "Datasource" and "GKE Datasource" select boxes to only display datasources valid for those fields. PTAL?

> *[Reviewable](https://reviewable.io/reviews/m-lab/prometheus-support/932#-:-N8_AZ5MAvFcS3zUlujw:b-i0uzz)* status: 0 of 1 approvals obtained

cristinaleonr (approved):
Great!
:lgtm: 

> *[Reviewable](https://reviewable.io/reviews/m-lab/prometheus-support/932#-:-N8_CTcxDRbRCabcEq0W:b-oqmnq5)* status: :shipit: complete! 1 of 1 approvals obtained

PR #939, opened 2022-08-12
Update GCS transfer alert to ignore archive-mlab-oti

stephen-soltesz:
Due to https://github.com/m-lab/dev-tracker/issues/729 archive files are not preserved in archive-mlab-oti bucket after transfer to the archive-measurement-lab bucket any more. And, the current alert `GCSTransfers_ArchiveFilesDoNotMatchOrMissing` asserts that the same number of files are in both archive-mlab-oti and archive-measurement-lab buckets. Because the new transfer policy violates the alert constraint, it fired: https://github.com/m-lab/dev-tracker/issues/740

This change updates the alert to a simpler `GCSTransfers_ArchiveFilesMissing` daily check for the presence of archive files in the archive-measurement-lab bucket. The `increase()` function will apply to all unique combinations of datatype labels. The `and` clause additionally filters datatypes that have no data transferred (e.g. older datatypes). With this alert, a false positive is possible if there are no new archives for a datatype for over a day. But, using our current metric history, this has not occurred in that last 16w.

- - -
This change is [Reviewable](https://reviewable.io/reviews/m-lab/prometheus-support/939)

### DIFF
--- a/config/federation/prometheus/alerts.yml
+++ b/config/federation/prometheus/alerts.yml
@@ -225,7 +225,7 @@ groups:
 #
 # The following alerts are based on the exact queries that mlab-ns runs to
 # determine the state of NDT services.
-# https://github.com/m-lab/mlab-ns/blob/master/server/mlabns/util/prometheus_status.py
+# https://github.com/m-lab/mlab-ns/blob/main/server/mlabns/util/prometheus_status.py
 #
 # The expression in the denominator of the queries can be read as:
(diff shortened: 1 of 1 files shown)


nkinkade (approved):
:lgtm:

> Reviewed 1 of 1 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/m-lab/prometheus-support/939#-:-N9Iar8x42BIyrsTSupg:bnfp4nl)* status: :shipit: complete! 1 of 1 approvals obtained

PR #948, opened 2022-08-26
Removes resource definitions for prometheus

nkinkade:
Just give Prometheus a sufficiently large VM, and let it have the resources it needs when it needs them. Predefining the resource requests and limits messes with the scheduler, and forces us to create a VM (possibly) larger than is actually required for Prometheus if it doesn't actually need that amount of resources.

Also, removes nodeSelector for various things. We don't need to be running stackdriver_exporter and kube-state-metrics on the Prometheus VM. Just let the scheduler run them on one of the nodes in the default pool. Also, adds a taint to the Prometheus VM to keep pods other than Prometheus off of it.

- - -
This change is [Reviewable](https://reviewable.io/reviews/m-lab/prometheus-support/948)


stephen-soltesz (approved):
As long as removing the node selectors don't prevent jobs from running where they're supposed to (i.e. crowding out resources on the static-ip node pool, or similar). :lgtm:

> Reviewed 8 of 8 files at r1.
> *[Reviewable](https://reviewable.io/reviews/m-lab/prometheus-support/948#-:-NAQuHYw8pgctVzt8xrw:bfx0v70)* status: :shipit: complete! 1 of 1 approvals obtained

nkinkade (commented):
The `static-outbound-ip` pool nodes have a taint to prevent anything other than what we want on them. I am also creating the prometheus "pool" (only one node) with a taint to prevent anything but prometheus on there. Everything else can run in the default pool, which has auto-scaling enabled (up to 6 nodes).

> *[Reviewable](https://reviewable.io/reviews/m-lab/prometheus-support/948#-:-NAQv1-g2BurWAgE53df:brzs062)* status: :shipit: complete! 1 of 1 approvals obtained

PR #956, opened 2022-10-17
Smoother predictions for disk utilization

stephen-soltesz:
Since prometheus appears to garbage collect its storage every week, the existing "Filesystem Available Estimate" panel includes unrealistic and confusing jumps during these garbage collection events.

This change updates the time range for the linear prediction to include 10d, to smooth out the large discontinuity. The result is that the prediction should be smoother and more representative of a real prediction.

- - -
This change is [Reviewable](https://reviewable.io/reviews/m-lab/prometheus-support/956)

### DIFF
--- a/config/federation/grafana/dashboards/Prometheus_SelfMonitoring.json
+++ b/config/federation/grafana/dashboards/Prometheus_SelfMonitoring.json
@@ -3,7 +3,10 @@
     "list": [
       {
         "builtIn": 1,
-        "datasource": "-- Grafana --",
+        "datasource": {
+          "type": "datasource",
+          "uid": "grafana"
+        },
         "enable": false,
         "hide": true,
         "iconColor": "rgba(0, 211, 255, 1)",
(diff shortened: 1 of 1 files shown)


nkinkade (approved):
:lgtm: with the caveat that I'm not currently able to see the result in sandbox, I believe, because branch `sandbox-soltesz-explicit-scrape-interval` was pushed after the commits in this PR, so the most recent build reflects those changes instead of these. But I assume you made the change, pushed to sandbox and verified that the panel was displaying more sensible results.

> Reviewed 1 of 1 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/m-lab/prometheus-support/956#-:-NElKfz8CqdYZ6Il71Ui:bbeqa7b)* status: :shipit: complete! 1 of 1 approvals obtained

stephen-soltesz:
Well, it looked good to me. If it doesn't to you, please let me know so that we can fix it.

PR #960, opened 2022-11-14
Changes table display value from "machine" to "node"

nkinkade:
For some metrics, the "machine" label value ends up being "prometheus-platform-cluster", which represents the machine where the exporter is running, and not the actual machine represented by the metric value. In this case, the "node" label should more reliably give us the value we are looking for.

The impetus for this PR came from the Ops Tactical meeting this morning where the SRE Overview dashboard was showing that the "prometheus-platform-cluster" node was in lame-duck mode, which doesn't make sense, and I had no recollection of having added a lame-duck taint to that node. Upon closer inspection, it turns out the lame-duck was actually for mlab1-tlv01, but for the above reason was display the Prom node. This PR should prevent that from happening again.

- - -
This change is [Reviewable](https://reviewable.io/reviews/m-lab/prometheus-support/960)

### DIFF
--- a/config/federation/grafana/dashboards/Ops_Tactical_SRE_Overview.json
+++ b/config/federation/grafana/dashboards/Ops_Tactical_SRE_Overview.json
@@ -3,7 +3,10 @@
     "list": [
       {
         "builtIn": 1,
-        "datasource": "-- Grafana --",
+        "datasource": {
+          "type": "datasource",
+          "uid": "grafana"
+        },
         "enable": false,
         "hide": true,
         "iconColor": "rgba(0, 211, 255, 1)",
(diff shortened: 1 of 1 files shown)


stephen-soltesz (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/m-lab/prometheus-support/960#-:-NGrsPT64pGH07roMCC2:bnfp4nl)* status: :shipit: complete! 1 of 1 approvals obtained

nkinkade (commented):
It shouldn't be, and I just did a test in sandbox, verifying that it behaves the same for both virtual and physical nodes.

> *[Reviewable](https://reviewable.io/reviews/m-lab/prometheus-support/960#-:-NGs-k1w2vQysVEUl68x:b-qenc3d)* status: :shipit: complete! 1 of 1 approvals obtained

PR #968, opened 2023-01-20
Add an alert for Locate V2 request errors

cristinaleonr:
This PR adds a new alert for the rate of Locate V2 request errors.
A description for the alert has also been added under https://github.com/m-lab/ops-tracker/wiki/Alerts-&-Troubleshooting#locate_toomanyerrors.

This change is [Reviewable](https://reviewable.io/reviews/m-lab/prometheus-support/968)

### DIFF
--- a/config/federation/prometheus/alerts.yml
+++ b/config/federation/prometheus/alerts.yml
@@ -37,6 +37,19 @@ groups:
       description: '{{ $labels.instance }} of job {{ $labels.job }} has been down
         for more than 10 minutes.'
 
+# TooManyProcessRestarts: a process has restarted more than twice in the last hour.
+  - alert: TooManyProcessRestarts
+    expr: resets(process_cpu_seconds_total[1h]) > 2
+    for: 10m
+    labels:
+      repo: dev-tracker
+      severity: ticket
+      cluster: prometheus-federation
+    annotations:
+      summary: Instance {{ $labels.instance }} has restarted more than twice in the
+        last hour.
+      description: https://github.com/m-lab/ops-tracker/wiki/Alerts-&-Troubleshooting#toomanyprocessrestarts
+
 ##
 ## SLOs
 ##
--- a/config/federation/prometheus/alerts.yml
+++ b/config/federation/prometheus/alerts.yml
@@ -568,19 +568,19 @@ groups:
       description: https://github.com/m-lab/ops-tracker/wiki/Alerts-&-Troubleshooting#mlabns_serverresponsemetricmissing
 
 # The alert Locate_TooManyErrors will fire when the rate of /nearest requests to the
-# Locate V2 is greater than 1% for 2 minutes. The locate_requests_total metrics come
+# Locate V2 is greater than 0.1% for 2 minutes. The locate_requests_total metrics come
 # from Prometheus.
   - alert: Locate_TooManyErrors
     expr: |
-      (sum (locate_requests_total{type="nearest", status!="OK"}) /
-      sum(locate_requests_total{type="nearest"})) > 0.01
+      (sum(rate(locate_requests_total{type="nearest", status!="OK"}[10m])/10) /
+      sum(rate(locate_requests_total{type="nearest"}[10m])/10)) > 0.001
     for: 2m
     labels:
       repo: dev-tracker
       severity: ticket
       cluster: prometheus-federation
     annotations:
-      summary: The rate of Locate V2 nearest request errors is greater than 1%.
+      summary: The rate of Locate V2 nearest request errors is greater than 0.1%.
       description: https://github.com/m-lab/ops-tracker/wiki/Alerts-&-Troubleshooting#locate_toomanyerrors
       dashboard: https://grafana.mlab-oti.measurementlab.net/d/8O9tInk4k/locate-service?orgId=1&refresh=5m&var-datasource=Prometheus%20%28mlab-oti%29&var-platformdatasource=Platform%20Cluster%20%28mlab-oti%29&var-bigquerydatasource=BigQuery%20%28mlab-oti%29&var-metro=All&var-experiment=ndt&viewPanel=4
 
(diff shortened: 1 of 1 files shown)


stephen-soltesz (commented):
> *[Reviewable](https://reviewable.io/reviews/m-lab/prometheus-support/968)* status: 0 of 1 approvals obtained (waiting on @cristinaleonr)

___
*[`config/federation/prometheus/alerts.yml` line 576 at r1](https://reviewable.io/reviews/m-lab/prometheus-support/968#-NMG4R5D46pZNHPV9vOv:-NMG4R5D46pZNHPV9vOw:b3bbzed) ([raw file](https://github.com/m-lab/prometheus-support/blob/7009e9bec553800d69dc7c766c95ae761b9ef8f8/config/federation/prometheus/alerts.yml#L576)):*
> ```YAML
>     expr: |
>       (sum (locate_requests_total{type="nearest", status!="OK"}) /
>       sum(locate_requests_total{type="nearest"})) > 0.01
> ```

What was the threshold you noticed manually? I'm inclined to believe that an even lower threshold should get our attention. At least 0.005 -- but I'm even wondering if 1/1000 would be wise.

And, perhaps an additional alert counting https://prometheus.io/docs/prometheus/latest/querying/functions/#resets ?

That could be a separate PR and can probably use a generic Go metric so any process we monitoring would get caught.

stephen-soltesz (commented):
> *[Reviewable](https://reviewable.io/reviews/m-lab/prometheus-support/968)* status: 0 of 1 approvals obtained (waiting on @cristinaleonr)

___
*[`config/federation/prometheus/alerts.yml` line 576 at r1](https://reviewable.io/reviews/m-lab/prometheus-support/968#-NMG4R5D46pZNHPV9vOv:-NMG9wOmBSzt5_P6Cfoc:b-un3csr) ([raw file](https://github.com/m-lab/prometheus-support/blob/7009e9bec553800d69dc7c766c95ae761b9ef8f8/config/federation/prometheus/alerts.yml#L576)):*

What was the threshold you noticed manually? I'm inclined to believe that an even lower threshold should get our attention. At least 0.005 -- but I'm even wondering if 1/1000 would be wise.

And, perhaps an additional alert counting https://prometheus.io/docs/prometheus/latest/querying/functions/#resets ?

That could be a separate PR and can probably use a generic Go metric so any process we monitoring would get caught.
</blockquote></details>

Also, I think this should use a `rate` calculation rather than the counter metric itself.

So, `sum(rate(...)) / sum(rate(....)) > X`

cristinaleonr (commented):
> *[Reviewable](https://reviewable.io/reviews/m-lab/prometheus-support/968)* status: 0 of 1 approvals obtained (waiting on @stephen-soltesz)

___
*[`config/federation/prometheus/alerts.yml` line 576 at r1](https://reviewable.io/reviews/m-lab/prometheus-support/968#-NMG4R5D46pZNHPV9vOv:-NMGDiEQA51qzrkmZOeb:b4qi19r) ([raw file](https://github.com/m-lab/prometheus-support/blob/7009e9bec553800d69dc7c766c95ae761b9ef8f8/config/federation/prometheus/alerts.yml#L576)):*

Also, I think this should use a `rate` calculation rather than the counter metric itself.

So, `sum(rate(...)) / sum(rate(....)) > X`
</blockquote></details>

That was just the threshold from the mlab-ns alert. I have changed it to 1/1000 and added a `rate`.

I will add the reset metric on Monday.

cristinaleonr (commented):
I added another alert for `resets()`. I used a threshold of > 2 per hour, but I'm open to suggestions.

> *[Reviewable](https://reviewable.io/reviews/m-lab/prometheus-support/968#-:-NMUfNYO4o-960BWE_O1:b-4vaygd)* status: 0 of 1 approvals obtained (waiting on @stephen-soltesz)

stephen-soltesz (approved):
Thank you for adding the wiki pages and linking to dashboards when available.

With the suggested change :lgtm:

> *[Reviewable](https://reviewable.io/reviews/m-lab/prometheus-support/968#-:-NMUp8wi9PyaOvC2f0BN:b-9ujdnx)* status: :shipit: complete! 1 of 1 approvals obtained (waiting on @cristinaleonr)

___
*[`config/federation/prometheus/alerts.yml` line 42 at r4](https://reviewable.io/reviews/m-lab/prometheus-support/968#-NMUo8gR2HCeGy-eMmi-:-NMUo8gR2HCeGy-eMmi0:by510xz) ([raw file](https://github.com/m-lab/prometheus-support/blob/9356afbb45cf8b30be0a974472419be6decf117e/config/federation/prometheus/alerts.yml#L42)):*
> ```YAML
> # TooManyProcessRestarts: a process has restarted more than twice in the last hour.
>   - alert: TooManyProcessRestarts
>     expr: resets(process_cpu_seconds_total[1h]) > 2
> ```

I agree finding the right threshold is tricky. This may end up being service dependent.

Our intention is to discover crash-loops - some restarts may be expected due to deployments or containers settling in a K8s context. 

To prevent false positives and to set a threshold that would have caught this recent event, I propose a higher threshold to start, say 6.

cristinaleonr (commented):
> *[Reviewable](https://reviewable.io/reviews/m-lab/prometheus-support/968)* status: :shipit: complete! 1 of 1 approvals obtained (waiting on @stephen-soltesz)

___
*[`config/federation/prometheus/alerts.yml` line 42 at r4](https://reviewable.io/reviews/m-lab/prometheus-support/968#-NMUo8gR2HCeGy-eMmi-:-NMV955hCT7Ko1VnT9GJ:b-896fix) ([raw file](https://github.com/m-lab/prometheus-support/blob/9356afbb45cf8b30be0a974472419be6decf117e/config/federation/prometheus/alerts.yml#L42)):*

I agree finding the right threshold is tricky. This may end up being service dependent.

Our intention is to discover crash-loops - some restarts may be expected due to deployments or containers settling in a K8s context. 

To prevent false positives and to set a threshold that would have caught this recent event, I propose a higher threshold to start, say 6.
</blockquote></details>

Done.

PR #972, opened 2023-01-31
Adds OOM alerts and new OOM panel on SRE Overview dashboard

nkinkade:
This new alert and dashboard panel are based on a new [log-based metric in each project](https://console.cloud.google.com/logs/metrics?project=mlab-sandbox), named "platform_cluster_oom". The Logging query for the log-based metric is:

```
resource.type="generic_node"
resource.labels.location="<project>"
jsonPayload.message:"Out of memory"
```

Log-based metrics are 1m samples that reflect the delta from the previous sample. For this reason, the new resources do not do any sort of aggregation or rate calculations, but instead just use the raw values returned by the metric (`stackdriver_generic_node_logging_googleapis_com_user_platform_cluster_oom`).

The alert is very sensitive, as OOM events shouldn't really every happen.

The new "OOM Events" panel on the SREOverview dashboard replaces the old, rather unhelpful panel "Kernel logs priority <=3", which we have never really used and the data was ill-defined.

- - -
This change is [Reviewable](https://reviewable.io/reviews/m-lab/prometheus-support/972)

### DIFF
--- a/config/federation/grafana/dashboards/Ops_Tactical_SRE_Overview.json
+++ b/config/federation/grafana/dashboards/Ops_Tactical_SRE_Overview.json
@@ -19,7 +19,7 @@
   "editable": true,
   "fiscalYearStartMonth": 0,
   "graphTooltip": 0,
-  "id": 76,
+  "id": 266,
   "links": [],
   "liveNow": false,
   "panels": [
@@ -627,9 +627,10 @@
       "dashLength": 10,
       "dashes": false,
       "datasource": {
+        "type": "prometheus",
         "uid": "$datasource"
       },
-      "description": "",
+      "description": "1m rate of OOM log messages for a node",
       "fieldConfig": {
         "defaults": {
           "links": []
@@ -680,17 +681,19 @@
           "datasource": {
             "uid": "$datasource"
           },
-          "expr": "sum by (log, priority) (stackdriver_generic_node_logging_googleapis_com_user_platform_cluster_kernel_logs{priority=~\"[0-3]\"})",
+          "editorMode": "code",
+          "expr": "rate(stackdriver_generic_node_logging_googleapis_com_user_platform_cluster_oom[5m]) * 60",
           "format": "time_series",
           "interval": "",
           "intervalFactor": 1,
           "legendFormat": "{{log}}",
+          "range": true,
           "refId": "A"
         }
       ],
       "thresholds": [],
       "timeRegions": [],
-      "title": "Kernel logs priority <=3",
+      "title": "OOM Events",
       "tooltip": {
         "shared": true,
         "sort": 2,
@@ -1181,7 +1184,7 @@
     "list": [
       {
         "current": {
-          "selected": false,
+          "selected": true,
           "text": "Prometheus (mlab-oti)",
           "value": "Prometheus (mlab-oti)"
         },
@@ -1200,7 +1203,7 @@
       },
       {
         "current": {
-          "selected": false,
+          "selected": true,
           "text": "Platform Cluster (mlab-oti)",
           "value": "Platform Cluster (mlab-oti)"
         },
@@ -1211,6 +1214,7 @@
         "name": "cluster",
         "options": [],
         "query": "prometheus",
+        "queryValue": "",
         "refresh": 1,
         "regex": "/Cluster/",
         "skipUrlSync": false,
@@ -1250,6 +1254,6 @@
   "timezone": "",
   "title": "Ops: Tactical & SRE Overview",
   "uid": "_fugwnWZk",
-  "version": 17,
+  "version": 49,
   "weekStart": ""
 }
--- a/config/federation/grafana/dashboards/Ops_Tactical_SRE_Overview.json
+++ b/config/federation/grafana/dashboards/Ops_Tactical_SRE_Overview.json
@@ -622,105 +622,137 @@
       "type": "table-old"
     },
     {
-      "aliasColors": {},
-      "bars": false,
-      "dashLength": 10,
-      "dashes": false,
       "datasource": {
         "type": "prometheus",
         "uid": "$datasource"
       },
-      "description": "1m rate of OOM log messages for a node",
+      "description": "",
       "fieldConfig": {
         "defaults": {
-          "links": []
+          "color": {
+            "mode": "thresholds"
+          },
+          "custom": {
+            "align": "left",
+            "displayMode": "auto",
+            "filterable": false,
+            "inspect": false
+          },
+          "links": [],
+          "mappings": [],
+          "min": 0,
+          "thresholds": {
+            "mode": "absolute",
+            "steps": [
+              {
+                "color": "green",
+                "value": null
+              },
+              {
+                "color": "red",
+                "value": 80
+              }
+            ]
+          },
+          "unit": "short"
         },
-        "overrides": []
+        "overrides": [
+          {
+            "matcher": {
+              "id": "byName",
+              "options": "node_id"
+            },
+            "properties": [
+              {
+                "id": "displayName",
+                "value": "Node"
+              }
+            ]
+          },
+          {
+            "matcher": {
+              "id": "byName",
+              "options": "Node"
+            },
+            "properties": [
+              {
+                "id": "custom.width",
+                "value": 345
+              }
+            ]
+          }
+        ]
       },
-      "fill": 1,
-      "fillGradient": 0,
       "gridPos": {
         "h": 9,
         "w": 6,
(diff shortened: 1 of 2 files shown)


stephen-soltesz (commented):
> *[Reviewable](https://reviewable.io/reviews/m-lab/prometheus-support/972)* status: 0 of 1 approvals obtained (waiting on @nkinkade)

___
*[`config/federation/prometheus/alerts.yml` line 1155 at r1](https://reviewable.io/reviews/m-lab/prometheus-support/972#-NN8cdBw4Of-OMkjfM1o:-NN8cdBw4Of-OMkjfM1p:b-sv69b9) ([raw file](https://github.com/m-lab/prometheus-support/blob/e86208389cdd78d5b8d3393794ed56f47bc0cb5f/config/federation/prometheus/alerts.yml#L1155)):*
> ```YAML
>       cluster: prometheus-federation
>     annotations:
>       summary: A node has experienced too many OOM events (more than 0)
> ```

Just to confirm that we're capturing the intent and describing the same thing.

The condition is `>1` so the minimum value will be 2 for the alert to fire. This description implies a single event should fire. I'm inclined to prefer `>1` to avoid any false positives. I imagined in the outage the other day there were significantly more than 1 right?

If the stackdriver metric only identifies events from a 1m window, perhaps a `sum_over_time(...[10m]) > 1` (for example) expression could allow us to ignore singleton events. (I see this type of expression is already in the dashboard).

What do you think?

nkinkade (commented):
> *[Reviewable](https://reviewable.io/reviews/m-lab/prometheus-support/972)* status: 0 of 1 approvals obtained (waiting on @stephen-soltesz)

___
*[`config/federation/prometheus/alerts.yml` line 1155 at r1](https://reviewable.io/reviews/m-lab/prometheus-support/972#-NN8cdBw4Of-OMkjfM1o:-NN8gH8dCBZzbzsDxM5o:bz62ty5) ([raw file](https://github.com/m-lab/prometheus-support/blob/e86208389cdd78d5b8d3393794ed56f47bc0cb5f/config/federation/prometheus/alerts.yml#L1155)):*

Just to confirm that we're capturing the intent and describing the same thing.

The condition is `>1` so the minimum value will be 2 for the alert to fire. This description implies a single event should fire. I'm inclined to prefer `>1` to avoid any false positives. I imagined in the outage the other day there were significantly more than 1 right?

If the stackdriver metric only identifies events from a 1m window, perhaps a `sum_over_time(...[10m]) > 1` (for example) expression could allow us to ignore singleton events. (I see this type of expression is already in the dashboard).

What do you think?
</blockquote></details>

Ah, good catch. My intention was indeed that it be `> 0`. In my mind, there are are probably no false positives with regard to a machine running out of memory and the oom-killer being invoked. It seems to me that if the oom-killer is invoked even a single time, then that is a problem, hence the current alert configuration.

I'm not opposed to relaxing the criteria, since the platform is self-healing (k8s) anyway. That said, wouldn't changing the expression to `sum_over_time(...[10m]) > 1` actually make the alert a bit more sensitive in a way? The expression, as it stands,  will only alert if the number of OOMs is greater than 1, for more than 2 minutes. This allows for the number to go up above 1 for at least one scrape cycle, then go down to 1 or 0 before the next scrape cycle begins. If we are summing over time, then there could be, say, 2 OOM events near the start of the 10m window, then perhaps the issue goes away and there are not more after that (a very brief resource spike). In that scenario, depending on timing, the alert might still fire, even though the underlying cause of the issue was brief and went away on its own.

Maybe the comparison should be something like `> 3`, or similar?

stephen-soltesz (commented):
> *[Reviewable](https://reviewable.io/reviews/m-lab/prometheus-support/972)* status: 0 of 1 approvals obtained (waiting on @nkinkade)

___
*[`config/federation/prometheus/alerts.yml` line 1155 at r1](https://reviewable.io/reviews/m-lab/prometheus-support/972#-NN8cdBw4Of-OMkjfM1o:-NN8lPOeAs7yUvWlsDfy:b-vg3lc6) ([raw file](https://github.com/m-lab/prometheus-support/blob/e86208389cdd78d5b8d3393794ed56f47bc0cb5f/config/federation/prometheus/alerts.yml#L1155)):*

Ah, good catch. My intention was indeed that it be `> 0`. In my mind, there are are probably no false positives with regard to a machine running out of memory and the oom-killer being invoked. It seems to me that if the oom-killer is invoked even a single time, then that is a problem, hence the current alert configuration.

I'm not opposed to relaxing the criteria, since the platform is self-healing (k8s) anyway. That said, wouldn't changing the expression to `sum_over_time(...[10m]) > 1` actually make the alert a bit more sensitive in a way? The expression, as it stands,  will only alert if the number of OOMs is greater than 1, for more than 2 minutes. This allows for the number to go up above 1 for at least one scrape cycle, then go down to 1 or 0 before the next scrape cycle begins. If we are summing over time, then there could be, say, 2 OOM events near the start of the 10m window, then perhaps the issue goes away and there are not more after that (a very brief resource spike). In that scenario, depending on timing, the alert might still fire, even though the underlying cause of the issue was brief and went away on its own.

Maybe the comparison should be something like `> 3`, or similar?
</blockquote></details>

My suggestion may be premature optimization - my goal is to reduce false positives or chatty alerts. My thinking is 2 events in N minutes is a more important signal for us to see than 1 event in 2 minutes. One might be transient, but 2 is worse. That said, I'm still very open to iterating and flexible about the threshold and time window. What do you recommend?

nkinkade (commented):
> *[Reviewable](https://reviewable.io/reviews/m-lab/prometheus-support/972)* status: 0 of 1 approvals obtained (waiting on @stephen-soltesz)

___
*[`config/federation/prometheus/alerts.yml` line 1155 at r1](https://reviewable.io/reviews/m-lab/prometheus-support/972#-NN8cdBw4Of-OMkjfM1o:-NN8pUi-7h2-5EgroAbj:b-3kh749) ([raw file](https://github.com/m-lab/prometheus-support/blob/e86208389cdd78d5b8d3393794ed56f47bc0cb5f/config/federation/prometheus/alerts.yml#L1155)):*

My suggestion may be premature optimization - my goal is to reduce false positives or chatty alerts. My thinking is 2 events in N minutes is a more important signal for us to see than 1 event in 2 minutes. One might be transient, but 2 is worse. That said, I'm still very open to iterating and flexible about the threshold and time window. What do you recommend?
</blockquote></details>

I changed the expression to `sum_over_time(...[5m]) > 1`. I don't know if that is the right setting, but it can be a start. The log-based metric I created in production yesterday around this time has not logged a single OOM event on the production platform, but I'll be curious to see how common these events are, or aren't.

stephen-soltesz (approved):
:lgtm: with a one request.

> *[Reviewable](https://reviewable.io/reviews/m-lab/prometheus-support/972#-:-NN8uBhp9sJ913jiEocC:bgd7xur)* status: :shipit: complete! 1 of 1 approvals obtained (waiting on @nkinkade)

___
*[`config/federation/prometheus/alerts.yml` line 1155 at r1](https://reviewable.io/reviews/m-lab/prometheus-support/972#-NN8cdBw4Of-OMkjfM1o:-NN8u4X48kJWXzUAgdkM:b6chlnn) ([raw file](https://github.com/m-lab/prometheus-support/blob/e86208389cdd78d5b8d3393794ed56f47bc0cb5f/config/federation/prometheus/alerts.yml#L1155)):*

I changed the expression to `sum_over_time(...[5m]) > 1`. I don't know if that is the right setting, but it can be a start. The log-based metric I created in production yesterday around this time has not logged a single OOM event on the production platform, but I'll be curious to see how common these events are, or aren't.  
</blockquote></details>

Thank you, me too!

The expression and the summary description are still out of sync. :)

nkinkade (commented):
> *[Reviewable](https://reviewable.io/reviews/m-lab/prometheus-support/972)* status: :shipit: complete! 1 of 1 approvals obtained (waiting on @stephen-soltesz)

___
*[`config/federation/prometheus/alerts.yml` line 1155 at r1](https://reviewable.io/reviews/m-lab/prometheus-support/972#-NN8cdBw4Of-OMkjfM1o:-NN8uyAr9Ai8ArWXkPNW:bmf678s) ([raw file](https://github.com/m-lab/prometheus-support/blob/e86208389cdd78d5b8d3393794ed56f47bc0cb5f/config/federation/prometheus/alerts.yml#L1155)):*

Thank you, me too!

The expression and the summary description are still out of sync. :) 
</blockquote></details>

Ah, comments are now synced with the expression. More precisely, I removed all specific numbers from the comments and summary to obviate the need to keep them in sync.

PR #973, opened 2023-02-17
Updates name of node pool for kube-ip

nkinkade:
We had a disk pressure issue on the static-public-ip nodes, which caused script-exporter to be evicted. It is not possible to resize the disks of an existing node pool. The only way to achieve this is to create a new node pool with the desired configurations and then migrate your workloads to this new node pool. I created a new node pool in mlab-oti named static-external-ip and migrated the workloads to this new node pool. This change configures kube-ip to be sure that those nodes get static IPv4 addresses. Meanwhile I have manually updated the kube-ip deployment in production with the change in this commit.

- - -
This change is [Reviewable](https://reviewable.io/reviews/m-lab/prometheus-support/973)

### DIFF
--- a/k8s/prometheus-federation/deployments/kubeip.yml
+++ b/k8s/prometheus-federation/deployments/kubeip.yml
@@ -32,7 +32,7 @@ spec:
             value: "prometheus-federation"
             # Nodepool to which KubeIP assigns addresses
           - name: "KUBEIP_NODEPOOL"
-            value: "static-public-ip"
+            value: "static-external-ip"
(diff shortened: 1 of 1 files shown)


cristinaleonr (approved):
:lgtm: 

> *[Reviewable](https://reviewable.io/reviews/m-lab/prometheus-support/973#-:-NOWdQ_b2vrx1URZGt-I:bg7igkv)* status: :shipit: complete! 1 of 1 approvals obtained

PR #974, opened 2023-02-27
Turn off debug logging for script-exporter

nkinkade:
Debug logging is only useful when you are debugging, and we are not debugging script-exporter. Furthermore, the log is appended to on every test, and eventually the logs get big and fill up the root filesystem.

- - -
This change is [Reviewable](https://reviewable.io/reviews/m-lab/prometheus-support/974)

### DIFF
--- a/k8s/prometheus-federation/deployments/script-exporter.yml
+++ b/k8s/prometheus-federation/deployments/script-exporter.yml
@@ -37,7 +37,7 @@ spec:
         - -config.file=/etc/script_exporter/script_exporter.yml
         env:
         - name: LOGX_DEBUG
-          value: 'false'
+          value: ''
         - name: MONITORING_SIGNER_KEY
           value: /keys/monitoring-signer-key.json
         - name: LOCATE_URL
--- a/k8s/prometheus-federation/deployments/script-exporter.yml
+++ b/k8s/prometheus-federation/deployments/script-exporter.yml
@@ -36,8 +36,6 @@ spec:
         args:
         - -config.file=/etc/script_exporter/script_exporter.yml
         env:
-        - name: LOGX_DEBUG
-          value: ''
         - name: MONITORING_SIGNER_KEY
           value: /keys/monitoring-signer-key.json
         - name: LOCATE_URL
--- a/k8s/prometheus-federation/deployments/script-exporter.yml
+++ b/k8s/prometheus-federation/deployments/script-exporter.yml
@@ -37,7 +37,7 @@ spec:
         - -config.file=/etc/script_exporter/script_exporter.yml
         env:
         - name: LOGX_DEBUG
-          value: 'true'
+          value: 'false'
         - name: MONITORING_SIGNER_KEY
           value: /keys/monitoring-signer-key.json
         - name: LOCATE_URL


stephen-soltesz (approved):
With suggested change, :lgtm:

> *[Reviewable](https://reviewable.io/reviews/m-lab/prometheus-support/974#-:-NPK3Ytv7GteIuCH3Bb8:bp6dxlu)* status: :shipit: complete! 1 of 1 approvals obtained (waiting on @nkinkade)

___
*[`k8s/prometheus-federation/deployments/script-exporter.yml` line 40 at r1](https://reviewable.io/reviews/m-lab/prometheus-support/974#-NPK3JRy-LfvJKWQPjTS:-NPK3JRy-LfvJKWQPjTT:b-gwbija) ([raw file](https://github.com/m-lab/prometheus-support/blob/f9c1243fef26d3ada19c48ff564e5b78b5fb2049/k8s/prometheus-federation/deployments/script-exporter.yml#L40)):*
> ```YAML
>         env:
>         - name: LOGX_DEBUG
>           value: ''
> ```

I'm not sure what an empty value does. This is passed as a flag value for `-logx.debug=` -- I recommend setting to `false` or removing the environment variable definition completely.

nkinkade (commented):
> *[Reviewable](https://reviewable.io/reviews/m-lab/prometheus-support/974)* status: :shipit: complete! 1 of 1 approvals obtained (waiting on @stephen-soltesz)

___
*[`k8s/prometheus-federation/deployments/script-exporter.yml` line 40 at r1](https://reviewable.io/reviews/m-lab/prometheus-support/974#-NPK3JRy-LfvJKWQPjTS:-NPK7Go7C78AMA8gBxO7:bapndm0) ([raw file](https://github.com/m-lab/prometheus-support/blob/f9c1243fef26d3ada19c48ff564e5b78b5fb2049/k8s/prometheus-federation/deployments/script-exporter.yml#L40)):*

I'm not sure what an empty value does. This is passed as a flag value for `-logx.debug=` -- I recommend setting to `false` or removing the environment variable definition completely.
</blockquote></details>

@stephen-soltesz: despite the odd name of the variable (making one think is designed to do something with the m-lab/go/logx package), I believe this environment variable is only evaluated in a shell script:

https://github.com/m-lab/script-exporter-support/blob/master/cache_exit_code.sh#L30

I can't see that the m-lab/go/logx package is imported by ndt5-client, so I don't think this environment variable would have any effect downstream of the cache_exit_code.sh script. 

I considered removing it altogether, but decided against it: 

https://github.com/m-lab/prometheus-support/commit/f9c1243fef26d3ada19c48ff564e5b78b5fb2049

In this case, the empty string just causes `[[ -n $LOGX_DEBUG ]]` to evaluate to false, so no debug logging happens.

What do you think?

stephen-soltesz (commented):
> *[Reviewable](https://reviewable.io/reviews/m-lab/prometheus-support/974)* status: :shipit: complete! 1 of 1 approvals obtained (waiting on @nkinkade)

___
*[`k8s/prometheus-federation/deployments/script-exporter.yml` line 40 at r1](https://reviewable.io/reviews/m-lab/prometheus-support/974#-NPK3JRy-LfvJKWQPjTS:-NPKAmak4lChfOpeI2VM:b-ybyiyz) ([raw file](https://github.com/m-lab/prometheus-support/blob/f9c1243fef26d3ada19c48ff564e5b78b5fb2049/k8s/prometheus-federation/deployments/script-exporter.yml#L40)):*

@stephen-soltesz: despite the odd name of the variable (making one think is designed to do something with the m-lab/go/logx package), I believe this environment variable is only evaluated in a shell script:

https://github.com/m-lab/script-exporter-support/blob/master/cache_exit_code.sh#L30

I can't see that the m-lab/go/logx package is imported by ndt5-client, so I don't think this environment variable would have any effect downstream of the cache_exit_code.sh script. 

I considered removing it altogether, but decided against it: 

https://github.com/m-lab/prometheus-support/commit/f9c1243fef26d3ada19c48ff564e5b78b5fb2049

In this case, the empty string just causes `[[ -n $LOGX_DEBUG ]]` to evaluate to false, so no debug logging happens.

What do you think?
</blockquote></details>

I'm confident it's also interpreted by the `monitoring-token` command too -https://github.com/m-lab/locate/blob/main/cmd/monitoring-token/monitoring-token.go#L57

I suspect the cache script is borrowing the environment variable for a similar purpose, but without considering that there could be other values (that's probably my fault).

For a quick fix, I'd recommend removing the variable entirely. And, for a longer fix, making `cache_exit_code.sh` act only on a value `true`.

nkinkade (commented):
> *[Reviewable](https://reviewable.io/reviews/m-lab/prometheus-support/974)* status: :shipit: complete! 1 of 1 approvals obtained (waiting on @stephen-soltesz)

___
*[`k8s/prometheus-federation/deployments/script-exporter.yml` line 40 at r1](https://reviewable.io/reviews/m-lab/prometheus-support/974#-NPK3JRy-LfvJKWQPjTS:-NPKDH6TAkLiJB4ddV-a:be2heku) ([raw file](https://github.com/m-lab/prometheus-support/blob/f9c1243fef26d3ada19c48ff564e5b78b5fb2049/k8s/prometheus-federation/deployments/script-exporter.yml#L40)):*

I'm confident it's also interpreted by the `monitoring-token` command too -https://github.com/m-lab/locate/blob/main/cmd/monitoring-token/monitoring-token.go#L57

I suspect the cache script is borrowing the environment variable for a similar purpose, but without considering that there could be other values (that's probably my fault).

For a quick fix, I'd recommend removing the variable entirely. And, for a longer fix, making `cache_exit_code.sh` act only on a value `true`.
</blockquote></details>

I sort of like leaving it there, since it make the configuration explicit, and is a reminder to anyone later on that debug logging can be enabled there. The logs that are the problem are the ones created by the shell script. I imagine any debug logging emitted by monitoring-token would be a k8s log, and automatically handled/truncated/rotated by the kubelet. But it's not a big issue one way or the other, so I just removed it.

PR #985, opened 2023-06-22
Add prometheus variable

stephen-soltesz:
This change adds a hidden `${prometheus}` datasource variable so the `site` and `metro` variable lookup can refer to a local instance of the prometheus datasource, which varies per project.

- - -
This change is [Reviewable](https://reviewable.io/reviews/m-lab/prometheus-support/985)

### DIFF
--- a/config/federation/grafana/dashboards/NDT_Metro_Performance_Distributions.json
+++ b/config/federation/grafana/dashboards/NDT_Metro_Performance_Distributions.json
@@ -24,7 +24,6 @@
   "editable": true,
   "fiscalYearStartMonth": 0,
   "graphTooltip": 0,
-  "id": 430,
   "links": [],
   "liveNow": false,
   "panels": [
@@ -89,8 +88,8 @@
             "autorange": true,
             "gridcolor": "#333",
             "range": [
-              -0.001046069847020934,
-              0.019875327093397743
+              -0.0010763954493664726,
+              0.020451513537962977
             ],
             "type": "linear"
           }
@@ -183,8 +182,8 @@
             "autorange": true,
             "gridcolor": "#333",
             "range": [
-              -0.00315602951411648,
-              0.05996456076821312
+              -0.0031723594122624824,
+              0.060274828832987154
             ],
             "type": "linear"
           }
@@ -277,8 +276,8 @@
             "autorange": true,
             "gridcolor": "#333",
             "range": [
-              -0.0005958825356060797,
-              0.011321768176515513
+              -0.0007333601562907096,
+              0.013933842969523482
             ],
             "type": "linear"
           }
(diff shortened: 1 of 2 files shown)


nkinkade (changes requested):
I have a general question about this approach. For other dashboards that require selecting an appropriate datasource, we provide a visible select box. In sandbox you can select the datasource from any project. In staging you can select staging and production datasources, and in production there is only one option (production), which is automatically selected. This strategy means that you can use the sandbox dashboard to view data from any project, and is sometimes good for exploration of all projects. This was a conscious design decision:

https://github.com/m-lab/prometheus-support/blob/main/apply-global-prometheus.sh#L108

> *[Reviewable](https://reviewable.io/reviews/m-lab/prometheus-support/985#-:-NYdB2CF-iX-o0yKR4eG:b-vdfwav)* status: 1 change requests, 0 of 1 approvals obtained (waiting on @stephen-soltesz)

___
*[`config/federation/grafana/dashboards/NDT_Metro_Performance_Distributions.json` line 482 at r1](https://reviewable.io/reviews/m-lab/prometheus-support/985#-NYdHNrI9RF89hqeBmrX:-NYdHNrI9RF89hqeBmrY:bgt1bvi) ([raw file](https://github.com/m-lab/prometheus-support/blob/b8107872cb84c0ab10ce790052e5be2280d6aa79/config/federation/grafana/dashboards/NDT_Metro_Performance_Distributions.json#L482)):*
> ```JSON
>         "query": "prometheus",
>         "refresh": 1,
>         "regex": "/Platform Cluster .*mlab-oti.*/",
> ```

Is it intended that this will select the production datasource in all projects?

stephen-soltesz (commented):
Your question is good. Since this datasource is only used to generate the list of sites or metros, I didn't want to expose an extra option to the user.

Ideally, the set of sites/metros could just come from the same BQ datasource selection. But, when I tried that, for some reason, the template variables are not filled in the way they are for actual queries. So, if I wanted to use the "${type}" variable to only return virtual nodes, that can't work (or I couldn't figure how).

While investigating why, I discovered that there is a new version of the bigquery datasource plugin.

* old -  https://github.com/doitintl/bigquery-grafana (retired Dec, 2022)
* new - https://github.com/grafana/google-bigquery-datasource

I don't know if the new version supports this. If it does then we could eliminate the additional datasource.

For now, I've added a second datasource selection. This is only used to get the set of sites or metros used to filter the BQ results used in the performance PDFs.

PTAL?

> *[Reviewable](https://reviewable.io/reviews/m-lab/prometheus-support/985#-:-NYdWrst8M5-PkK2Kxgo:b-7gxoxa)* status: 1 change requests, 0 of 1 approvals obtained

nkinkade (approved):
:lgtm:

> Reviewed 2 of 2 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/m-lab/prometheus-support/985#-:-NYdxICrEY8UrCs3wngb:bnfp4nl)* status: :shipit: complete! 1 of 1 approvals obtained (waiting on @stephen-soltesz)

PR #988, opened 2023-07-25
Use datasource variables throughout

stephen-soltesz:
Evidently, the GCP billing dashboard has never worked outside of the sandbox environment. This was due to selecting explicit datasources that were serialized as "uid" strings that are unique per-Grafana instance. So the uid in sandbox for a BigQuery datasource is different from the uid in staging or oti, even if the target project is the same (e.g. BigQuery in mlab-oti).

This change adds two project datasource variables for bqSource and promSource and uses them for every panel. This change pins the project to mlab-oti since this is where the billing data is located and for the one prometheus metric, we only care about the mlab-oti data volume.

### DIFF
--- a/config/federation/grafana/dashboards/GCP_Billing.json
+++ b/config/federation/grafana/dashboards/GCP_Billing.json
@@ -71,7 +71,7 @@
     {
       "datasource": {
         "type": "doitintl-bigquery-datasource",
-        "uid": "0zzHsf77z"
+        "uid": "${bqSource}"
       },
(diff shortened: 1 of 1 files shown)


cristinaleonr (approved):
:lgtm: 

> *[Reviewable](https://reviewable.io/reviews/m-lab/prometheus-support/988#-:-NaD6l87EDx1nhi-Teb-:bg7igkv)* status: :shipit: complete! 1 of 1 approvals obtained

PR #994, opened 2023-08-11
Fix BigQuery exporter GCP Billing query

stephen-soltesz:
Since the revision to the `billing.unified` view to prevent future breakage due to schema changes in the underlying export tables, we removed a column used by the original bq_billing_hourly.sql query, and this query has been broken. However, these metrics are not actively used and obviated by the availability of the BigQuery datasource plugin.

So, this change simplifies the query significantly to export only daily, 30day and annual average costs. This query is run every three hours. And, can be the input to a future set of alerts on excess usage.

- - -
This change is [Reviewable](https://reviewable.io/reviews/m-lab/prometheus-support/994)


robertodauria (approved):
:lgtm:

> Reviewed 4 of 4 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/m-lab/prometheus-support/994#-:-NbnykmS9hob093Og8cK:bnfp4nl)* status: :shipit: complete! 1 of 1 approvals obtained

PR #999, opened 2023-08-21
Move from doitintl BigQuery plugin to the Grafana official one

robertodauria:
The `doitintl-bigquery-datasource` plugin [was deprecated](https://github.com/doitintl/bigquery-grafana) in Dec 2022. This PR installs the official Grafana BigQuery plugin and sets up datasources for every project. 

A subsequent PR will migrate the dashboards that are currently using the doitintl plugin to this new one.

- - -
This change is [Reviewable](https://reviewable.io/reviews/m-lab/prometheus-support/999)

PR #1001, opened 2023-08-23
Update Locate Service dashboard to use the new BQ datasource plugin

robertodauria:
This PR migrates the Locate Service monitoring dashboard to the new BQ datasource plugin (#999). It's important to note that the `$__timeFilter(date)` previously used in the query does not work with grafana-bigquery-datasource if `date` is of type `DATE`. In such cases, it must be replaced by an equivalent filter using the `$__from` and `$__to` macros (returning the timestamp as milliseconds). For example:

```WHERE date BETWEEN EXTRACT(date FROM TIMESTAMP_MILLIS($__from)) AND EXTRACT(date FROM TIMESTAMP_MILLIS($__to))```

This PR may also serve as an example of how to migrate other dashboards in the future.

- - -
This change is [Reviewable](https://reviewable.io/reviews/m-lab/prometheus-support/1001)

### DIFF
--- a/config/federation/grafana/dashboards/Locate_Service.json
+++ b/config/federation/grafana/dashboards/Locate_Service.json
@@ -442,7 +442,7 @@
       "targets": [
         {
           "datasource": {
-            "type": "doitintl-bigquery-datasource",
+            "type": "grafana-bigquery-datasource",
             "uid": "${bigquerydatasource}"
           },
           "editorMode": "code",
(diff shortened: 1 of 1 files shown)


cristinaleonr (approved):
:lgtm: 

> *[Reviewable](https://reviewable.io/reviews/m-lab/prometheus-support/1001#-:-NcY5g7ADbM4XixiPoLq:bg7igkv)* status: :shipit: complete! 1 of 1 approvals obtained

PR #1004, opened 2023-08-28
Enable nginx-ingress and oauth2 authentication for data-pipeline Prometheus

robertodauria:
This PR replicates the configuration we have in the prometheus-federation cluster in the data-pipeline cluster:
- nginx-ingress with a static external IP address
- cert-manager to obtain and renew TLS certificates automatically
- a TLS-only ingress for prometheus (running on a default port) + an internal ClusterIP service


apply-cluster.sh
```diff
@@ -16,6 +16,10 @@ USAGE="PROJECT=<projectid> CLUSTER=<cluster> $0"
 PROJECT=${PROJECT:?Please provide project id: $USAGE}
 CLUSTER=${CLUSTER:?Please provide cluster name: $USAGE}
 
+# Version numbers for Helm and Helm charts.
+K8S_HELM_VERSION="v3.3.0"
+K8S_INGRESS_NGINX_VERSION="4.2.1"
```
nkinkade, line 21: We define these exact variables and values in [apply-global-prometheus.sh](https://github.com/m-lab/prometheus-support/blob/main/apply-global-prometheus.sh#L27). Might we consider creating a separate file for variables of this type that both apply scripts import to avoid duplication? It seems there could be other things in the variables file like `OAUTH_PROXY_CLIENT.*` variables?

apply-cluster.sh
```diff
@@ -26,6 +30,54 @@ kubectl create configmap prometheus-cluster-config \
...
+sed -i -e 's|{{OAUTH_PROXY_CLIENT_ID}}|'${!OAUTH_PROXY_CLIENT_ID}'|g' \
+    -e 's|{{OAUTH_PROXY_CLIENT_SECRET}}|'${!OAUTH_PROXY_CLIENT_SECRET}'|g' \
+    -e 's|{{OAUTH_PROXY_COOKIE_SECRET}}|'${!OAUTH_PROXY_COOKIE_SECRET}'|g' \
+    k8s/data-pipeline/deployments/oauth2-proxy.yml
+
+# Additional k8s resources installed via Helm
+#
+# Download Helm
+curl -O https://get.helm.sh/helm-${K8S_HELM_VERSION}-linux-amd64.tar.gz
+tar -zxvf helm-${K8S_HELM_VERSION}-linux-amd64.tar.gz
```
nkinkade, line 48: I don't think there is a need for the `v` flag to tar here, which may just pollute the build logs.

apply-cluster.sh
```diff
@@ -26,6 +30,54 @@ kubectl create configmap prometheus-cluster-config \
...
+
+# Add repos
+./linux-amd64/helm repo add ingress-nginx https://kubernetes.github.io/ingress-nginx
+#./linux-amd64/helm repo add jetstack https://charts.jetstack.io
```
nkinkade, line 52: Does this jetstack repo need to be uncommented for the cert-manager install to work?

helm/data-pipeline/ingress-nginx/mlab-sandbox.yml
```diff
@@ -0,0 +1,5 @@
+controller:
+  service:
+    loadBalancerIP: 34.27.69.115
```
nkinkade, line 3: Was this IP address reserved manually by you? All new infrastructure should be deployed by Terraform.  Could we add this IP definition to the data-pipeline module in terraform-support?

https://github.com/m-lab/terraform-support/tree/main/modules/data-pipeline

Maybe add it to the data-pipeline.tf file?

k8s/data-pipeline/deployments/oauth2-proxy.yml
```diff
@@ -0,0 +1,36 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  labels:
+    run: oauth2-proxy
```
nkinkade, line 5: Does this label need to be here? I believe all the label selector stuff for the deployment is already handled under `spec`.

k8s/data-pipeline/services/oauth2-proxy.yml
```diff
@@ -0,0 +1,15 @@
+apiVersion: v1
+kind: Service
+metadata:
+  labels:
+    app: oauth2-proxy
```
nkinkade, line 5: Similar to above, is this label necessary here?

nkinkade (changes requested):
@robertodauria: I'm not sure what happened with my review from Monday on Reviewable.io. All my comments are in the Reviewable review, but when I click publish I get a `PERMISSION_DENIED` error for some reason. No matter what I do, I can't seem to get past that. I'm not sure whether that this the same error I hit on Monday without noticing, or whether I failed to publish on Monday and the error has something to do with the delay?

In any case, I'm doing this review directly on Github because Reviewable is wedged for this particular review.

I've added a few comments. I also notice that good number of files don't end with a NL. Not a big deal, but both Github and Reviewable are pretty noisy about it.

apply-cluster.sh
robertodauria: I moved some of the common varables to `config.sh` and now both scripts source it. Thoughts?

apply-cluster.sh
robertodauria: Removed (this command is now in `cloudbuild.yaml`)

apply-cluster.sh
robertodauria: Yes. (Moved to `cloudbuild.yaml`)

helm/data-pipeline/ingress-nginx/mlab-sandbox.yml
robertodauria: I'll open a separate PR in terraform-support. Would you prefer holding on approving this one until we have this IP (and the ones I need to add to staging and prod) fully managed by terraform?

k8s/data-pipeline/deployments/oauth2-proxy.yml
robertodauria: I think you're right.

k8s/data-pipeline/services/oauth2-proxy.yml
robertodauria: Removed.

nkinkade (approved):
:lgtm:, with a few questions that are not blocking but just for my own information.

> Reviewed 9 of 12 files at r1, 4 of 5 files at r2, 1 of 1 files at r3, 2 of 2 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/m-lab/prometheus-support/1004#-:-Ncy2oa4Di3nUm8RPJW3:b6p9a8c)* status: :shipit: complete! 1 of 1 approvals obtained (waiting on @robertodauria)

___
*[`config/cluster/prometheus/prometheus.yml` line 0 at r4](https://reviewable.io/reviews/m-lab/prometheus-support/1004#-NdC46sK3rS7WZInA9OU:-NdC46sK3rS7WZInA9OV:bjxhu71) ([raw file](https://github.com/m-lab/prometheus-support/blob/3e86ed5f3e766922ee1aee0fc55b83b402f306b1/config/cluster/prometheus/prometheus.yml#L0)):*
To be sure I understand, this is the same file as the former prometheus.yml.template file, with the only difference being that the `{{CLUSTER}}` value is now hard-coded. Why not just a rename of that file with that one small modification.
___
*[`helm/data-pipeline/ingress-nginx/mlab-sandbox.yml` line 3 at r1](https://reviewable.io/reviews/m-lab/prometheus-support/1004#gh-1310525852:-NdC21Wa8J7p52oomLmp:b-rd0lme) ([raw file](https://github.com/m-lab/prometheus-support/blob/4c4e62f71f67e558ad5ea6559ec9bfc74dd52cdb/helm/data-pipeline/ingress-nginx/mlab-sandbox.yml#L3)):*
<details><summary><i>Previously, robertodauria (Roberto D'Auria) wrote…</i></summary><blockquote>

I'll open a separate PR in terraform-support. Would you prefer holding on approving this one until we have this IP (and the ones I need to add to staging and prod) fully managed by terraform? 
</blockquote></details>

No, we don't need to block this PR on a separate one on terraform-support. The IP can stay the same, it just needs to be defined in TF and imported into the current state.
___
*[`k8s/data-pipeline/deployments/oauth2-proxy.yml` line 5 at r1](https://reviewable.io/reviews/m-lab/prometheus-support/1004#gh-1310527191:-NdC2LzwCHBzpl09ideS:bzfhw50) ([raw file](https://github.com/m-lab/prometheus-support/blob/4c4e62f71f67e558ad5ea6559ec9bfc74dd52cdb/k8s/data-pipeline/deployments/oauth2-proxy.yml#L5)):*
<details><summary><i>Previously, robertodauria (Roberto D'Auria) wrote…</i></summary><blockquote>

I think you're right.
</blockquote></details>

Tiny nit: it looks like you removed the actual label, but have left an empty `labels:` map? If so, just don't forget to remove `labels:`.
___
*[`k8s/data-pipeline/services/prometheus-tls-service.yml` line 16 at r4](https://reviewable.io/reviews/m-lab/prometheus-support/1004#-NdC3ei-2ZUb1iJ6lH5Y:-NdC3ei-2ZUb1iJ6lH5Z:b-btl2h) ([raw file](https://github.com/m-lab/prometheus-support/blob/3e86ed5f3e766922ee1aee0fc55b83b402f306b1/k8s/data-pipeline/services/prometheus-tls-service.yml#L16)):*
> ```YAML
>     run: prometheus-server
>   sessionAffinity: None
>   type: NodePort
> ```

Just for my own information, which node will be entrypoint for this service? My understanding of NodePort is that all nodes in the cluster will listen on a particular port, and will forward traffic to a suitable service endpoint.
___
*[`apply-cluster.sh` line 21 at r1](https://reviewable.io/reviews/m-lab/prometheus-support/1004#gh-1310524250:-NdC11sP4pVSAdv6VBnl:b7iiv5g) ([raw file](https://github.com/m-lab/prometheus-support/blob/4c4e62f71f67e558ad5ea6559ec9bfc74dd52cdb/apply-cluster.sh#L21)):*
<details><summary><i>Previously, robertodauria (Roberto D'Auria) wrote…</i></summary><blockquote>

I moved some of the common varables to `config.sh` and now both scripts source it. Thoughts?
</blockquote></details>

I like it. It's one more file and one more level of indirection, but it avoids duplication and variables in sync between two files.
___
*[`k8s/data-pipeline/services/prometheus-service.yml` line 0 at r4](https://reviewable.io/reviews/m-lab/prometheus-support/1004#-NdC3-myEI7ZSbBOX53X:-NdC3-mz3vDS2dlEmi1N:bht16i5) ([raw file](https://github.com/m-lab/prometheus-support/blob/3e86ed5f3e766922ee1aee0fc55b83b402f306b1/k8s/data-pipeline/services/prometheus-service.yml#L0)):*
Out of curiosity, whey the filename change? Without understanding the name change better, it looks to me like the filename now has a redundant "service" in the name i.e., 

services/prometheus-**service**.yml

robertodauria (commented):
> *[Reviewable](https://reviewable.io/reviews/m-lab/prometheus-support/1004)* status: :shipit: complete! 1 of 1 approvals obtained (waiting on @nkinkade)

___
*[`k8s/data-pipeline/services/prometheus-tls-service.yml` line 16 at r4](https://reviewable.io/reviews/m-lab/prometheus-support/1004#-NdC3ei-2ZUb1iJ6lH5Y:-NdCSJOXBfB8nLZ-D9hU:b1o784t) ([raw file](https://github.com/m-lab/prometheus-support/blob/3e86ed5f3e766922ee1aee0fc55b83b402f306b1/k8s/data-pipeline/services/prometheus-tls-service.yml#L16)):*

Just for my own information, which node will be entrypoint for this service? My understanding of NodePort is that all nodes in the cluster will listen on a particular port, and will forward traffic to a suitable service endpoint.
</blockquote></details>

The deployment matching `run: prometheus-server` is https://github.com/m-lab/prometheus-support/blob/main/k8s/data-pipeline/deployments/prometheus.yml#L23 which only runs on a node with:

```
      nodeSelector:
        prometheus-node: 'true'
```

This matches a single node, which is the one dedicated to Prometheus.
___
*[`k8s/data-pipeline/services/prometheus-service.yml` line  at r4](https://reviewable.io/reviews/m-lab/prometheus-support/1004#-NdC3-myEI7ZSbBOX53X:-NdCRaQFC6tg6HPp9_d_:b-897we8) ([raw file](https://github.com/m-lab/prometheus-support/blob/3e86ed5f3e766922ee1aee0fc55b83b402f306b1/k8s/data-pipeline/services/prometheus-service.yml#L)):*

Out of curiosity, whey the filename change? Without understanding the name change better, it looks to me like the filename now has a redundant "service" in the name i.e., 

services/prometheus-**service**.yml
</blockquote></details>

Mostly for consistency with prometheus-federation. I think the reason why prometheus-federation used this naming scheme is that `prometheus-tls-oauth` actually contains ingresses, not services. The `-service` suffix identifies the two actual services. I agree it's confusing.

robertodauria (commented):
> *[Reviewable](https://reviewable.io/reviews/m-lab/prometheus-support/1004)* status: :shipit: complete! 1 of 1 approvals obtained (waiting on @nkinkade)

___
*[`config/cluster/prometheus/prometheus.yml` line  at r4](https://reviewable.io/reviews/m-lab/prometheus-support/1004#-NdC46sK3rS7WZInA9OU:-NdCT1jh-9n20_f_4HZO:bpsqaqv) ([raw file](https://github.com/m-lab/prometheus-support/blob/3e86ed5f3e766922ee1aee0fc55b83b402f306b1/config/cluster/prometheus/prometheus.yml#L)):*

To be sure I understand, this is the same file as the former prometheus.yml.template file, with the only difference being that the `{{CLUSTER}}` value is now hard-coded. Why not just a rename of that file with that one small modification.
</blockquote></details>

This had been generated by the deployment script while I was testing locally. :confused:  I've removed it from this changeset.

robertodauria (commented):
Following your observations on resource naming scheme and the usage of NodePort, I've made a few changes:
- The resources' names should make more sense now
- The prometheus service has type ClusterIP now (everything still works since access is mediated by ingress-nginx).

PTAL again? :slightly_smiling_face: 

> *[Reviewable](https://reviewable.io/reviews/m-lab/prometheus-support/1004#-:-NdF_E1nDqqv7ungRH66:b-pqjvrf)* status: :shipit: complete! 1 of 1 approvals obtained (waiting on @nkinkade)

___
*[`k8s/data-pipeline/services/prometheus-tls-service.yml` line 16 at r4](https://reviewable.io/reviews/m-lab/prometheus-support/1004#-NdC3ei-2ZUb1iJ6lH5Y:-NdF_ADSAOXVtg3QzvlF:b-y7i8mk) ([raw file](https://github.com/m-lab/prometheus-support/blob/3e86ed5f3e766922ee1aee0fc55b83b402f306b1/k8s/data-pipeline/services/prometheus-tls-service.yml#L16)):*
<details><summary><i>Previously, robertodauria (Roberto D'Auria) wrote…</i></summary><blockquote>

The deployment matching `run: prometheus-server` is https://github.com/m-lab/prometheus-support/blob/main/k8s/data-pipeline/deployments/prometheus.yml#L23 which only runs on a node with:

```
      nodeSelector:
        prometheus-node: 'true'
```

This matches a single node, which is the one dedicated to Prometheus.
</blockquote></details>

I've changed this to be `ClusterIP` (and renamed it)

nkinkade (approved):
:lgtm:, thanks!

> Reviewed 2 of 2 files at r5, 7 of 7 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/m-lab/prometheus-support/1004#-:-NdGc_eR8YW_yndPkwLx:bpqxd5f)* status: :shipit: complete! 1 of 1 approvals obtained

PR #1007, opened 2023-08-30
Removes ContainerLogsMissingInStackdriver alert

nkinkade:
We recently stopped pushing experiment logs to GCP:

https://github.com/m-lab/k8s-support/pull/849

Vector should still be pushing logs from some containers like node-exporter, cAdvisor, flannel, etc. However, it would appear that these container produce very few logs and trigger this alert.

Vector is also still pushing kernel logs over above a certain level. Again, it seems that a good chunk of machines do not push kernel logs frequently enough to make a reliable alert.

All this said, we have observed that Vector sometimes gets into an error state with regard to GCP, which has previously only been visible because of the alert being removed by this commit. We need to figure out a way to alert when Vector cannot push logs to Stackdriver, but does not exit or produce anything but a log message. Maybe Vector can push its own logs to Stackdriver, and we can search for error messages in a custom log-based metric?

- - -
This change is [Reviewable](https://reviewable.io/reviews/m-lab/prometheus-support/1007)


stephen-soltesz (approved):
:lgtm:

Does vector not report any metrics of it's own?

If there is no metric for auth failures, that could be an upstream contribution given how frequently it occurs. And, we could alert on that. It is a hazard for us to depend on something (even if just kernel logs) if we do not know whether it is running.

> *[Reviewable](https://reviewable.io/reviews/m-lab/prometheus-support/1007#-:-Nd7bPMO5hUdQw18CMb8:bybdm0b)* status: :shipit: complete! 1 of 1 approvals obtained

PR #1008, opened 2023-08-31
Panel to show amount of data transferred

cristinaleonr:
This PR adds a new panel to the Early Exit dashboard displaying the amount of data transferred per machine type.

[Panel](https://grafana.mlab-sandbox.measurementlab.net/d/W8JPPzzIz/ndt-early-exit?orgId=1&viewPanel=6). 

[Dashboard](https://grafana.mlab-sandbox.measurementlab.net/d/W8JPPzzIz/ndt-early-exit?orgId=1).

- - -
This change is [Reviewable](https://reviewable.io/reviews/m-lab/prometheus-support/1008)


cristinaleonr (commented):
Thanks for the feedback! I'm happy to chat over GVC if that is easier. That said, the only new panel was the "Amount of Data Transferred" one.

> nit: Please use consistent style within the query for tables using the backticks.

Done.

> While it will look funny at first, I recommend splitting curves by totals "with early\_exit" and "without early exit".

I added a separate panel with the breakdown with/without early exit. I left the aggregated one since @phillipa requested it.

> I'm wondering if we could do a normalized PDF chart with two lines with and without early exit parameter.

Let's talk this through and add it in another PR.

> nit: I recommend a second `WITH` clause rather than the nested select for readability.

Done.

> Since the other panels contrast with and without, I recommend including a "without" or "total" line here also.

Done.

> *[Reviewable](https://reviewable.io/reviews/m-lab/prometheus-support/1008#-:-NdasJg6ELwvglJTBTYJ:b-2hd4uq)* status: 0 of 1 approvals obtained

stephen-soltesz (approved):
I'm sorry -- I realized after the fact that I commented on more than you were expecting.

Your updates are first rate. Thank you.

:lgtm:

> *[Reviewable](https://reviewable.io/reviews/m-lab/prometheus-support/1008#-:-Ndaut5HCh9sDiOWb6Vs:bd4gbnz)* status: :shipit: complete! 1 of 1 approvals obtained

PR #1019, opened 2023-11-29
Upgrade to Grafana v10.2.2 & fix broken dashboards due to upgrade

nkinkade:
For the most part, the only dashboards that were broken were ones relying on the doitintl BigQuery plugin and/or the ae3e Plotly Panel plugin. There were a few other small breakages that I took care of too.

NOTE: I _only_ modified/fixed provisioned dashboards. Any adhoc dashboards in sandbox that people have created and use regularly may still be broken. The most common fixes are:

* Modify any datasource variable to use the newer Google BigQuery plugin from Grafana.
* For BigQuery panels, make sure that "Format" for each query panel has an appropriate value (e.g., Table, Time Series).
* In the "Script" section of Plotly Panel configurations remove the `.buffered` field from data sources, since it no longer exists.

- - -
This change is [Reviewable](https://reviewable.io/reviews/m-lab/prometheus-support/1019)

### DIFF
--- a/config/federation/grafana/dashboards/Alert_DownloaderIsFailingToUpdate.json
+++ b/config/federation/grafana/dashboards/Alert_DownloaderIsFailingToUpdate.json
@@ -4,7 +4,10 @@
       {
         "$$hashKey": "object:2035",
         "builtIn": 1,
-        "datasource": "-- Grafana --",
+        "datasource": {
+          "type": "datasource",
+          "uid": "grafana"
+        },
         "enable": false,
         "hide": true,
         "iconColor": "rgba(0, 211, 255, 1)",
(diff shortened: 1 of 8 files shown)


nkinkade (commented):
I generally did not attempt to fix dashboards or panels that are also broken in mlab-oti, since my only intention was to fix breakage specifically caused by the upgrade to Grafana v10.2.2.  Indeed, I probably shouldn't have touched the Alert_DownloaderIsFailingToUpdate dashboard at all, since the dashboard is 100% broken in mlab-oti, but that was the first one I looked at and I noticed it was trying to use datasources like "Data Proc", which don't exist any longer. So the dashboard in sandbox is working 50% better than the one in production:

https://grafana.mlab-oti.measurementlab.net/d/ZGuYht1mk/alert-downloaderisfailingtoupdate?orgId=1

Same with the NDT_CanaryValidation dashboard. It doesn't work in production because, at the very least, it is trying to query some table that no longer exists. What I tried to do for this dashboard in sandbox was to at least make sure that the BigQuery datasources were working at set correctly, and that panels had all the right attributes, otherwise I left it broken, as in production:

https://grafana.mlab-oti.measurementlab.net/d/kFrLGgLVu/ndt-canary-validation?orgId=1

> *[Reviewable](https://reviewable.io/reviews/m-lab/prometheus-support/1019#-:-NkW49c19v3zC_1w-CD2:b3t5kvi)* status: 0 of 1 approvals obtained

cristinaleonr (approved):
Fair point. :lgtm: 

> *[Reviewable](https://reviewable.io/reviews/m-lab/prometheus-support/1019#-:-NkW66WV53hsjun6fC10:b-lg408n)* status: :shipit: complete! 1 of 1 approvals obtained

PR #1035, opened 2024-03-20
Changes downloader alert to fire after 72h, not 21h

nkinkade:
The downloader used to download databases about every 8h. We have reconfigured it to only download about every 24h. Additionally, there is no need for this alert to be super sensitive. It is okay if annotations happen with databases that are only a couple days old. However, if downloader can't download all the files for more than 3d, then fire an alert.

- - -
This change is [Reviewable](https://reviewable.io/reviews/m-lab/prometheus-support/1035)

### DIFF
--- a/config/federation/prometheus/alerts.yml
+++ b/config/federation/prometheus/alerts.yml
@@ -74,21 +74,18 @@ groups:
         the switch itself, or with the transit provider.
       dashboard: 'https://grafana.mlab-oti.measurementlab.net/d/SuqnZ6Hiz/?orgId=1&var-site_name={{$labels.site}}'
 
-# DownloaderIsFailingToUpdate: The downloader hasn't successfully retrieved the files in
-# at least 21 hours, meaning that at least the last two download attempts have failed.
+# DownloaderIsFailingToUpdate: The downloader hasn't successfully retrieved
+# all databases in more than 72 hours.
   - alert: DownloaderIsFailingToUpdate
-    expr: time() - downloader_last_success_time_seconds > (21 * 60 * 60)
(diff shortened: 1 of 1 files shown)


cristinaleonr (approved):
I think 3 days is a bit on the longer end to find out if the downloader is failing.

:lgtm: 

> *[Reviewable](https://reviewable.io/reviews/m-lab/prometheus-support/1035#-:-NtRtX4wELpsV7Mjag48:b-mhfh31)* status: <span>:shipit:</span> complete! 1 of 1 approvals obtained

nkinkade (commented):
My reasoning was that downloading every 24h +/- 4h, means that two download cycles might occur only every 54 hours, or just a little shy of 2.5 days. Failing one download cycle should be tolerated because transient errors occurs. And it seemed to me that a single database failing to download even every two download cycles should also probably be tolerated, for both transient errors or possible problems with the database providers that could last a day or two. So 72h was somewhat of a nice round number that would squarely encompass two download cycles, plus a little extra time. Another consideration was that MaxMind only updates the GeoLite2 databases twice a week, so there is no need to alert on a MaxMind failure more than every few days, since the data didn't change in that time anyway. It appears that CAIDA updates its routeviews databases every day, but I would be surprised if those change dramatically from day to day. The notion was that it's probably okay to annotate using data that could be only very _slightly_ stale.

> *[Reviewable](https://reviewable.io/reviews/m-lab/prometheus-support/1035#-:-NtRu2I29CYpGWc7xe1y:b-hgbrq4)* status: <span>:shipit:</span> complete! 1 of 1 approvals obtained

cristinaleonr:
If you're back here, it would be helpful to add this reasoning as a comment.

PR #1043, opened 2024-06-02
Modify Locate_TooManyErrors threshold

cristinaleonr:
This PR modifies the `Locate_TooManyErrors` to have an error threshold of 0.5%. The previous threshold (0.1%) was too sensitive.

- - -
This change is [Reviewable](https://reviewable.io/reviews/m-lab/prometheus-support/1043)

### DIFF
--- a/config/federation/prometheus/alerts.yml
+++ b/config/federation/prometheus/alerts.yml
@@ -582,7 +582,7 @@ groups:
 # from Prometheus.
   - alert: Locate_TooManyErrors
     expr: |
-      (sum(rate(locate_requests_total{type="nearest", status!~"OK|Too Many Requests", condition!~"client location"}[2m]))) /
+      (sum(rate(locate_requests_total{type="nearest", status!~"OK|Too Many Requests"}[2m]))) /
       sum(rate(locate_requests_total{type="nearest"}[2m])) > 0.005
     for: 2m
     labels:
--- a/config/federation/prometheus/alerts.yml
+++ b/config/federation/prometheus/alerts.yml
@@ -578,12 +578,12 @@ groups:
       description: https://github.com/m-lab/ops-tracker/wiki/Alerts-&-Troubleshooting#mlabns_serverresponsemetricmissing
 
 # The alert Locate_TooManyErrors will fire when the rate of /nearest requests to the
-# Locate V2 is greater than 0.1% for 2 minutes. The locate_requests_total metrics come
+# Locate V2 is greater than 0.5% for 2 minutes. The locate_requests_total metrics come
 # from Prometheus.
   - alert: Locate_TooManyErrors
     expr: |
       (sum(rate(locate_requests_total{type="nearest", status!~"OK|Too Many Requests", condition!~"client location"}[2m]))) /
-      sum(rate(locate_requests_total{type="nearest"}[2m])) > 0.001
+      sum(rate(locate_requests_total{type="nearest"}[2m])) > 0.005
     for: 2m
     labels:
       repo: dev-tracker
@@ -591,25 +591,9 @@ groups:
       cluster: prometheus-federation
       page_project: mlab-oti
     annotations:
-      summary: The rate of Locate V2 nearest request errors is greater than 0.1%.
+      summary: The rate of Locate V2 nearest request errors is greater than 0.5%.
       description: https://github.com/m-lab/ops-tracker/wiki/Alerts-&-Troubleshooting#locate_toomanyerrors
       dashboard: https://grafana.mlab-oti.measurementlab.net/d/8O9tInk4k/locate-service?orgId=1&refresh=5m&var-datasource=Prometheus%20%28mlab-oti%29&var-platformdatasource=Platform%20Cluster%20%28mlab-oti%29&var-bigquerydatasource=BigQuery%20%28mlab-oti%29&var-metro=All&var-experiment=ndt&viewPanel=4
-
-# The alert Locate_TooManyClientLocationErrors will fire when the rate of /nearest requests to the
-# Locate V2 with client location errors is greater than 1% for 2 minutes.
-  - alert: Locate_TooManyClientLocationErrors
-    expr: |
-      (sum(rate(locate_requests_total{type="nearest", condition="client location"}[2m]))) /
-      sum(rate(locate_requests_total{type="nearest"}[2m])) > 0.01
-    for: 2m
-    labels:
-      repo: dev-tracker
-      severity: ticket
-      cluster: prometheus-federation
-      page_project: mlab-oti
-    annotations:
-      summary: The rate of Locate V2 client location errors is greater than 1%.
-      description: https://github.com/m-lab/ops-tracker/wiki/Alerts-&-Troubleshooting#locate_toomanyclientlocationerrors
 
 # The alert Locate_HighLantecy will fire when the latency of at least 2% of /nearest requests
 # is greater than 1 second. The locate_request_handler_duration metrics come from Prometheus.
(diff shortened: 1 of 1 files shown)


nkinkade (changes requested):
> *[Reviewable](https://reviewable.io/reviews/m-lab/prometheus-support/1043)* status: 1 change requests, 0 of 1 approvals obtained (waiting on @cristinaleonr)

___
*[`config/federation/prometheus/alerts.yml` line 585 at r1](https://reviewable.io/reviews/m-lab/prometheus-support/1043#-NzUN5cc-Dk7mK0jdjqR:-NzUN5cc-Dk7mK0jdjqS:b-ywfi9k) ([raw file](https://github.com/m-lab/prometheus-support/blob/53bc8aa5b0590513669623e4fc64593b4c26e141/config/federation/prometheus/alerts.yml#L585)):*
> ```YAML
>   - alert: Locate_TooManyErrors
>     expr: |
>       (sum(rate(locate_requests_total{type="nearest", status!~"OK|Too Many Requests", condition!~"client location"}[2m]))) /
> ```

My only concern here is that omitting "client location" conditions could mask a more serious, generalized error condition that could affect many clients. I don't know how it could happen specifically, but say there is some bug in our Maxmind code, or a change to the Maxmind schema catches us off guard and breaks all Maxmind look ups. Does this seem possible to you, or am I being too cautious? I just want be sure that we don't believe that "client location" errors could ever be more systemic and where we would definitely want a page.

Tiny nit: the match on condition probably doesn't need to be a regular expression.
___
*[`config/federation/prometheus/alerts.yml` line 612 at r1](https://reviewable.io/reviews/m-lab/prometheus-support/1043#-NzUNdOT1z5si6PGT_DJ:-NzUNdOT1z5si6PGT_DK:b-qzvwm8) ([raw file](https://github.com/m-lab/prometheus-support/blob/53bc8aa5b0590513669623e4fc64593b4c26e141/config/federation/prometheus/alerts.yml#L612)):*
> ```YAML
>     annotations:
>       summary: The rate of Locate V2 client location errors is greater than 1%.
>       description: https://github.com/m-lab/ops-tracker/wiki/Alerts-&-Troubleshooting#locate_toomanyclientlocationerrors
> ```

Thank you for documenting the alert!

cristinaleonr (commented):
> *[Reviewable](https://reviewable.io/reviews/m-lab/prometheus-support/1043)* status: 1 change requests, 0 of 1 approvals obtained (waiting on @nkinkade)

___
*[`config/federation/prometheus/alerts.yml` line 585 at r1](https://reviewable.io/reviews/m-lab/prometheus-support/1043#-NzUN5cc-Dk7mK0jdjqR:-NzUOyZx6YqjoXO-7xDI:b-6hztjc) ([raw file](https://github.com/m-lab/prometheus-support/blob/53bc8aa5b0590513669623e4fc64593b4c26e141/config/federation/prometheus/alerts.yml#L585)):*

My only concern here is that omitting "client location" conditions could mask a more serious, generalized error condition that could affect many clients. I don't know how it could happen specifically, but say there is some bug in our Maxmind code, or a change to the Maxmind schema catches us off guard and breaks all Maxmind look ups. Does this seem possible to you, or am I being too cautious? I just want be sure that we don't believe that "client location" errors could ever be more systemic and where we would definitely want a page.

Tiny nit: the match on condition probably doesn't need to be a regular expression.
</blockquote></details>

This is definitely possible, so we still want to alert on this condition. That said, it's not something we could directly fix (we don't own MaxMind), so I reasoned a compromise would be to move the condition to a less sensitive, non-pageable alert. We can adjust either of these knobs depending on our level of comfort with the scenario you describe, but the alternative (leaving it as-is) is that we continue to get pages :sweat_smile:.

nkinkade (changes requested):
> *[Reviewable](https://reviewable.io/reviews/m-lab/prometheus-support/1043)* status: 1 change requests, 0 of 1 approvals obtained (waiting on @cristinaleonr)

___
*[`config/federation/prometheus/alerts.yml` line 585 at r1](https://reviewable.io/reviews/m-lab/prometheus-support/1043#-NzUN5cc-Dk7mK0jdjqR:-NzV-3U0AsTGThNH2jb1:br904di) ([raw file](https://github.com/m-lab/prometheus-support/blob/53bc8aa5b0590513669623e4fc64593b4c26e141/config/federation/prometheus/alerts.yml#L585)):*
<details><summary><i>Previously, cristinaleonr (Cristina Leon) wrote…</i></summary><blockquote>

This is definitely possible, so we still want to alert on this condition. That said, it's not something we could directly fix (we don't own MaxMind), so I reasoned a compromise would be to move the condition to a less sensitive, non-pageable alert. We can adjust either of these knobs depending on our level of comfort with the scenario you describe, but the alternative (leaving it as-is) is that we continue to get pages :sweat_smile:.
</blockquote></details>

Just thinking out loud: would another reasonable alternative be to raise the threshold for the alert? Right now the alert threshold is a 0.1% error rate.  If only 1.1 in a 1000 requests fails for more than 2m the alert will fire. The current threshold seems to have served us pretty well for the past ~1.5 years. However, it is _pretty_ sensitive. Looking back at the alerts that fired over the weekend, none of the error rates appear to have gone over 0.175%. Would it be reasonable to tolerate an error of, say, 0.0025 or similar. Honestly, it seems we could probably even raise it to something like .005 and be pretty safe. 

One other thought would be to maybe have two alerts with the same criteria, but with two different thresholds. One higher threshold to trigger pages, and another lower one that would trigger a non-pageable alert we could safely investigate during normal business hours. I'm just hesitant to categorically split out "client location" errors into a separate non-pageable alert, in the event that some malfunction within our control triggers that specific error condition.

What do you think?

nkinkade (approved):
LGTM, after you remove the `condition!~"client location"` condition in the Locate_TooMayErrors alert.

Thanks!

> Reviewed 1 of 1 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/m-lab/prometheus-support/1043#-:-NzV7_y96E7GMA-59sBW:bcysiqb)* status: <span>:shipit:</span> complete! 1 of 1 approvals obtained (waiting on @cristinaleonr)

cristinaleonr (commented):
Thanks!

> *[Reviewable](https://reviewable.io/reviews/m-lab/prometheus-support/1043#-:-NzV8MS-1QeN1QENaaWD:b-ucfknh)* status: <span>:shipit:</span> complete! 1 of 1 approvals obtained (waiting on @nkinkade)

___
*[`config/federation/prometheus/alerts.yml` line 585 at r1](https://reviewable.io/reviews/m-lab/prometheus-support/1043#-NzUN5cc-Dk7mK0jdjqR:-NzV6ybp7j_kEsIg23Or:bnnatx5) ([raw file](https://github.com/m-lab/prometheus-support/blob/53bc8aa5b0590513669623e4fc64593b4c26e141/config/federation/prometheus/alerts.yml#L585)):*

Just thinking out loud: would another reasonable alternative be to raise the threshold for the alert? Right now the alert threshold is a 0.1% error rate.  If only 1.1 in a 1000 requests fails for more than 2m the alert will fire. The current threshold seems to have served us pretty well for the past ~1.5 years. However, it is _pretty_ sensitive. Looking back at the alerts that fired over the weekend, none of the error rates appear to have gone over 0.175%. Would it be reasonable to tolerate an error of, say, 0.0025 or similar. Honestly, it seems we could probably even raise it to something like .005 and be pretty safe. 

One other thought would be to maybe have two alerts with the same criteria, but with two different thresholds. One higher threshold to trigger pages, and another lower one that would trigger a non-pageable alert we could safely investigate during normal business hours. I'm just hesitant to categorically split out "client location" errors into a separate non-pageable alert, in the event that some malfunction within our control triggers that specific error condition.

What do you think?
</blockquote></details>

That makes sense. I considered doing that at first, but was afraid to miss Locate-originated errors by relaxing the threshold. That said, 0.5% seems reasonable.

PR #1049, opened 2024-08-01
Fully removes doitintl BigQuery datasource references

nkinkade:
The deprecated doitintl BigQuery datasource is still referenced in a number of dashboards, and the datasources themselves also still exist, even if not used. This PR fully removes all references to this deprecated datasource, as well as adds them to the `deleteDatasources` list so that Grafana will delete these provisioned datasources.

Also:
* took this opportunity to upgrade both Grafana and Prometheus to the latest versions
* upgraded the Ops_SwitchOverview dashboard to use the newer "Timeseries" panels, instead of the deprecated "Graph (old)" panel types. 
* gets rid of provisioning/datasources/deleted.yml, in favor of a file named provisioning/datasources/config.yml. The new file contains the same things as the old deleted.yml, but also includes the new-ish feature of `prune: true`, which is _supposed_ to automatically delete provisioned datasources when the provisioning file that created it goes away.
* removes the deprecated `[alerting]` section from grafana.ini, which is superseded by `[unified_alerting]`

- - -
This change is [Reviewable](https://reviewable.io/reviews/m-lab/prometheus-support/1049)


cristinaleonr (approved):
:lgtm: 

> *[Reviewable](https://reviewable.io/reviews/m-lab/prometheus-support/1049#-:-O3YJeKZA059RVTiSnRj:bg7igkv)* status: <span>:shipit:</span> complete! 1 of 1 approvals obtained

PR #1051, opened 2024-08-19
Removes process count panels from two dashboards

nkinkade:
Related to https://github.com/m-lab/k8s-support/pull/900

I have never really used the process count panel of the Site Overview dashboard, and I believe that the "K8s: Topk Machine & Pod Resources" dashboard is very old and unused. In any case, removing these the process count panel from these dashboards doesn't really change the base intent of either.

- - -
This change is [Reviewable](https://reviewable.io/reviews/m-lab/prometheus-support/1051)

### DIFF
--- a/config/federation/grafana/dashboards/K8s_SiteOverview.json
+++ b/config/federation/grafana/dashboards/K8s_SiteOverview.json
@@ -24,7 +24,7 @@
   "editable": true,
   "fiscalYearStartMonth": 0,
   "graphTooltip": 1,
-  "id": 246,
+  "id": 65,
   "links": [],
   "liveNow": false,
   "panels": [
@@ -114,13 +114,13 @@
         "y": 1
       },
       "id": 40,
-      "links": [],
       "maxDataPoints": 100,
       "options": {
         "colorMode": "value",
         "graphMode": "none",
         "justifyMode": "auto",
         "orientation": "horizontal",
+        "percentChangeColorMode": "standard",
         "reduceOptions": {
           "calcs": [
             "lastNotNull"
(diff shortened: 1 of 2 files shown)


stephen-soltesz (approved):
:lgtm:

I don't object to eliminating this if node-exporter is causing too much trouble.

I'm also sad to see this metric being lost - process count / running processes is such a fundamental metric that has been a signal to the root cause of some past issues. Maybe it's less of a concern today. Strange that it's so difficult to get such a simple metric.

> *[Reviewable](https://reviewable.io/reviews/m-lab/prometheus-support/1051#-:-O4gmNrPDDd_4Y-R92kV:bngv2v0)* status: <span>:shipit:</span> complete! 1 of 1 approvals obtained

nkinkade (commented):
I agree it is a nice-to-have, just in case. But when I started weighing figuring out how to modify the default containerd AppArmor profile (which is apparently dynamically loaded into the kernel at run time, and isn't even on disk) vs simply disabling a collector that I hadn't personally made use of in perhaps many years, the better solution was to remove the collector. Additionally, the "processes" collector is [disabled by default in node_exporter](https://github.com/prometheus/node_exporter?tab=readme-ov-file#disabled-by-default), which means that the authors consider possibly detrimental to the system or TSDB for any number of reasons.

If for some reason we find that we need those metrics in the future, adding the collector back is trivial. 

> *[Reviewable](https://reviewable.io/reviews/m-lab/prometheus-support/1051#-:-O4gokK4-WlbE-v6zGzo:b-j3d0ye)* status: <span>:shipit:</span> complete! 1 of 1 approvals obtained

PR #1054, opened 2024-08-26
Set mlab-oti as the default for the Pipeline dashboard

cristinaleonr:
This PR updates the default data source in the ["Pipeline: SLIs"](https://grafana.mlab-sandbox.measurementlab.net/d/q4MrNzh7k/pipeline3a-slis?orgId=1) dashboard to be mlab-oti. I was investigating an alert earlier and ended up accidentally looking at sandbox data, when the majority of the time these dashboards are looked at is for production insights.

The time range is also updated from `now-2d` to `now-7d` to get a wider perspective.

The number of changes looks a little more significant only because the dashboard hasn't been updated in a long time.

- - -
This change is [Reviewable](https://reviewable.io/reviews/m-lab/prometheus-support/1054)

### DIFF
--- a/config/federation/grafana/dashboards/Pipeline_SLIs.json
+++ b/config/federation/grafana/dashboards/Pipeline_SLIs.json
@@ -3,7 +3,10 @@
     "list": [
       {
         "builtIn": 1,
-        "datasource": "-- Grafana --",
+        "datasource": {
+          "type": "datasource",
+          "uid": "grafana"
+        },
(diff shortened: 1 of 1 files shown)


stephen-soltesz (approved):
> *[Reviewable](https://reviewable.io/reviews/m-lab/prometheus-support/1054)* status: <span>:shipit:</span> complete! 1 of 1 approvals obtained

PR #1057, opened 2024-09-05
Add prometheus deployment to autojoin cluster

stephen-soltesz:
This change adds support for deployment of a prometheus configuration to the autojoin cluster.

This change required several changes to remain compatible with the data-pipeline cluster configuration.

- `apply-data-cluster.sh` was once generic and is now generic again, renamed to `apply-cluster.sh`
- Add a new `autojoin` cluster directory, e.g. k8s/autojoin
- Update these cluster configurations to use `monitoring-node: 'true'` node selector - data-pipeline clusters have been manually updated to include this label on the prometheus nodepools.
- Remove the prometheus-federation configuration for scraping autonode metrics.
- Performs conditional deployments in mlab-oti or mlab-autojoin
```
$ host prometheus-autojoin.mlab-autojoin.measurementlab.net
prometheus-autojoin.mlab-autojoin.measurementlab.net has address 34.30.73.176
```

For example:

- https://prometheus-autojoin.mlab-sandbox.measurementlab.net/graph?g0.expr=lame_duck_experiment%7Bdeployment%3D%22byos%22%7D&g0.tab=1&g0.display_mode=lines&g0.show_exemplars=0&g0.range_input=1h

- - -
This change is [Reviewable](https://reviewable.io/reviews/m-lab/prometheus-support/1057)


stephen-soltesz:
@nkinkade this is a draft PR - but I wanted to ask you to take a look to see if you see it going in an okay direction or not - in particular it does feel like I'm making up a template convention - I wonder if helm for prometheus might be a better strategy for the "cluster" configuration - the autojoin deployment needs an additional prometheus configuration for scraping the byos nodes. Any thoughts are welcomed

nkinkade (commented):
This repo is more complicated than, say, k8s-support, because it deploys to multiple clusters. The layout of the k8s-support repository was loosely based on this one with a structure that is generally like:

./k8s/<resource>/<name>.yaml
./config/<name>.yaml
./helm/<override>.yaml

I think it could be useful if this repo had the same structure, but just inserting `<cluster>` after each top level directory... and having names that always match. For example, though I know it pretty well today, it has always been a slight source of confusion that in this repository we have `./k8s/prometheus-federation` and `./config/federation`, which refer to the same cluster, but have different directory names. 

I haven't thought too much about this, but having a single "./config/cluster" directory that basically contains a Prometheus template config used by multiple clusters might make the layout less easy to understand without digging into shell scripts.  Having separate directories for each cluster in each top level directory might make for more files and more duplication, but might be easier to understand. Something like:

./k8s/<cluster>/<resource>/<name>.yaml
./config/<cluster>/<name>.yaml
./helm/<cluster>/<override>.yaml

I guess it's a tradeoff, because if multiple clusters all need/want the exact same Prometheus configuration, then it's hard to justify maintaining several copies of the config that must all be kept in sync over time. On the other hand, having separate configurations is easier to look at, and allows us to modify the Prometheus config for each cluster.  I see that the Prom config template has only been touched a couple times in the past 4 years. Could sym links be used?

There are random thoughts, and may not actually make sense, or even be feasible for reasons I'm not thinking of. I also don't address the possibility of using helm to install Prometheus.

> *[Reviewable](https://reviewable.io/reviews/m-lab/prometheus-support/1057#-:-O63JKSJ5KGqGs_dVC7R:b5jrl6u)* status: 0 of 1 approvals obtained

nkinkade (approved):
LGTM

> Reviewed 15 of 43 files at r1, 28 of 28 files at r2, 1 of 1 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/m-lab/prometheus-support/1057#-:-O6N-4Vb8YU4ejUpyOs5:bb6ywmx)* status: <span>:shipit:</span> complete! 1 of 1 approvals obtained

PR #1063, opened 2024-10-16
Adds an alert for when GCE costs are >50% of montly average

nkinkade:
As an example, the current daily average for the past 30 days is $11,318. On October 9th, due to the Exactly Labs bug, the Compute Engine costs were $16,965. Using >50% as the threshold would have _just_ caught the Exactly Labs bug (11318 * 1.5 = 16977). Lowering the threshold to >25% would have triggered a false alarm on a few occasions over the past month due to normal variation in Compute Engine costs. We could probably get away with maybe a >40% threshold, but >50% seemed like a nice round number and will catch the more serious increases after a couple days.

- - -
This change is [Reviewable](https://reviewable.io/reviews/m-lab/prometheus-support/1063)

### DIFF
--- a/config/federation/prometheus/alerts.yml
+++ b/config/federation/prometheus/alerts.yml
@@ -1352,3 +1352,24 @@ groups:
       summary: Daily BigQuery costs are 2x the average BigQuery costs for the month.
       description: https://github.com/m-lab/ops-tracker/wiki/Alerts-&-Troubleshooting#billing_dailybigqueryincrease
       dashboard: https://grafana.mlab-oti.measurementlab.net/d/f4efc343-661f-4899-a00f-a4ffeb294e5d/bigquery-usage?orgId=1&from=now-7d&to=now
+
+# Billing_DailyGCEIncrease fires when the daily Compute Engine costs for any of
+# the last 3 days is 50% over the average Compute Engine costs for the last 30
+# days. Because the GCP billing export may take over a day to complete, the
+# alert checks as far as 2 days back to get an accurate measure of the daily
+# costs. That said, even if only with partial data, we opportunistically also
+# check today's and yesterday's cost.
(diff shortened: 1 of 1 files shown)


stephen-soltesz (commented):
Would 40% have caused any false positives? I'm imagining a 40% increase, which is still pretty significant. Alternately, what is the lowest threshold that doesn't generate any false positives in the last 6mo?

Thank you for adding the wiki page summarizing this new alert.

> *[Reviewable](https://reviewable.io/reviews/m-lab/prometheus-support/1063#-:-O9MoDbCE4XBClvP7bUU:b-bbcgu0)* status: 0 of 1 approvals obtained

nkinkade (commented):
I averaged the daily Compute Engine cost from June 1 through October 15, which ended up being right at $11,000. I only went back to June 1 because that is a few days after I changed virtual probability to 50% for VMs colocated in metros with at least one physical site. After that date daily costs went down.

140%  of $11,000 ($15,400) would have caused one false positive back in July, where the daily cost was $15,721 for some reason. 145% only caught the Exactly Labs incident on Oct 9. 

I have modified the percent to 145%, which isn't a big change, but at least a little more sensitive.

What do you think?

> *[Reviewable](https://reviewable.io/reviews/m-lab/prometheus-support/1063#-:-O9QN5eZ0rk7ejEhinqs:bibpl1p)* status: 0 of 1 approvals obtained

stephen-soltesz (approved):
:lgtm:

> Reviewed 1 of 1 files at r2.
> *[Reviewable](https://reviewable.io/reviews/m-lab/prometheus-support/1063#-:-O9QilMZ9CVQnhzRw3x1:bnfp4nl)* status: <span>:shipit:</span> complete! 1 of 1 approvals obtained

PR #1075, opened 2025-02-27
Adds "fqdn" label to script_exporter e2e metrics + new alert

nkinkade:
The Locate Service filters on label "fqdn" when looking to see whether e2e testing to a machine is failing. This PR adds that label Autojoin e2e testing so that Locate can effectively stop serving a machine where e2e testing is not working.

Additionally, in anticipation of moving M-Lab-managed VMs to leveraging the Autojoin system, this PR adds a new alert that will fire when Autojoin is less than 10d away from removing a DNS entry for an M-Lab-managed VM. This should give operators plenty of lead time to resolve the issue. In any case, by the time this alert may fire, other k8s alerts for the node should have already fired.

- - -
This change is [Reviewable](https://reviewable.io/reviews/m-lab/prometheus-support/1075)

### DIFF
--- a/config/federation/prometheus/alerts.yml
+++ b/config/federation/prometheus/alerts.yml
@@ -1381,11 +1381,16 @@ groups:
 # Autojoin system will delete a DNS record for an autojoin client if it hasn't
 # heard from that client more than 14d. Autojoin clients are supposed to
 # checkin every hour. This alert conservatively allows an M-Lab autojoined
-# machine to not checkin for 4 days before firing.
+# machine to not checkin for 4 days before firing. Since this alert is only
+# meant to uncover issues with the Autojoin client successfully registering, or
+# some bug in the Autojoin API where client registrations are not properly
+# updated in Memorystore, we only trigger this alert when script_exporter E2E
+# tests are succeeding. If E2E test are failing, then there is some other issue
+# and k8s alerts should cover them.
   - alert: AutojoinDNSRecordTooCloseToExpiring
     expr: |
       ((autojoin_dns_expiration - time()) / 86400) < 10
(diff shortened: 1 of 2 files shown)


nkinkade:
@robertodauria: I pushed another commit to this PR after I noticed alerts in sandbox for VMs that no longer exist. This is part of the larger problem where the Autojoin garbage collection TTL is 14d, which leaves around metrics for a bunch of VMs that no longer actually exist. In this case, the alert rightfully fires because Autojoin is going to remove the DNS entry for the machine in less than 10 days, because it hasn't checked in, because the machine no longer exists. The question is how we alert when a DNS entry might be removed for VM that actually still exists.

Please check my logic on this commit. At first it may not seem intuitive that the alert will only fire when E2E tests to ndt-server are *working*, since failing E2E tests seem like a problem and could possibly be related to why the machine is failing to register. However, my logic is that this new alert is meant to uncover issues only with registration, either in the client or API and not problems with the ndt-server container. If E2E testing is happening in the first place then it means that the machine successfully registered at least once, which would trigger E2E testing to start, and ndt-server should be running at that point, regardless of whether future registration attempts started failing in some way. That said, this does not necessarily feel ideal.

Thoughts?

robertodauria (approved):
:lgtm: 
This approach makes sense to me. The certificate's expiration is only relevant if the node in question is actually serving NDT tests.

> *[Reviewable](https://reviewable.io/reviews/m-lab/prometheus-support/1075#-:-OKS7gld2RfyZ_-13pK1:bcwxzwc)* status: <span>:shipit:</span> complete! 1 of 1 approvals obtained

PR #1077, opened 2025-03-07
Improve monitoring for the autojoin cluster

nkinkade:
This an unreasonably large PR. I got carried away. The gist is that this improves monitoring for the autojoin cluster. Some highlights:

* Adds a new alert for when the Autojoin API is returning too many request errors. The alert is generated by the autojoin instance of Prometheus, but is processed by the Alertmanager running in the prometheus-federation cluster. We need to determine whether the alert condition are right/good. We also need to create some additional alerts, but the framework is now set up.
* Moves e2e script-exporter monitoring of autonodes out of prometheus-federation and into the autojoin cluster. With these changes the prometheus-federation cluster scrapes the *script_success* metrics from the autojoin cluster so that Locate will get the metrics and can properly stop serving and autonode when the e2e testing is failing.
* Adds blackbox_exporter monitoring of autonodes for all ports we care about, which includes ndt-server and "sidecars". 
* Improves the Autojoin:SiteOverview dashboard by adding a new panel named "Port Status" that indicates whether all expected ports are reachable.

- - -
This change is [Reviewable](https://reviewable.io/reviews/m-lab/prometheus-support/1077)


robertodauria (approved):
:lgtm:

> Reviewed 18 of 18 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/m-lab/prometheus-support/1077#-:-OL-P01W4NGl-P53jwzi:bnfp4nl)* status: <span>:shipit:</span> complete! 1 of 1 approvals obtained

PR #1078, opened 2025-03-24
Adds new alert for when Jostler data volume is too low in autojoin

nkinkade:
This is almost identical to a simlar alert we have in k8s-support to monitor Pusher data volume:

https://github.com/m-lab/k8s-support/blob/main/config/prometheus/alerts.yml#L582

I don't know if 70% is the right threshold here, but it can serve as a start. When I ran the query manually in production, it did correctly show that the alert would have fired right after the incident began. What I don't know yet is whether 70% is too sensitive for the byos platform.

- - -
This change is [Reviewable](https://reviewable.io/reviews/m-lab/prometheus-support/1078)

### DIFF
--- a/config/autojoin/prometheus/alerts.yml
+++ b/config/autojoin/prometheus/alerts.yml
@@ -76,10 +76,10 @@ groups:
     expr: |
       datatype:jostler_bytes_per_bundle:increase24h
         < (0.7 * quantile by(datatype)(0.5,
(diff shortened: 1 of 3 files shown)


robertodauria (approved):
:lgtm: - we can always change it later if it's too sensitive.

> Reviewed 3 of 3 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/m-lab/prometheus-support/1078#-:-OM9s9mYBCTEpQJrLa1x:b-9ykbw)* status: <span>:shipit:</span> complete! 1 of 1 approvals obtained

PR #1083, opened 2025-08-26
Fixes a small bug in the Autojoin_SiteOverview dashboard

nkinkade:
The panels were not correctly filtering on org name, so if for any reason the same machine name belonged to two orgs, then the panel would not display properly. This shouldn't normally happen, but, for example, I accidentally created an org with the wrong name, the ISP brought up the machine, then I realized my mistake and created a new org with the right name, and the ISP brought that one up. We then had metrics for the same machine name, but two different orgs. For some reason this bug is affecting the "mlab" org too, and this should fix that also.

https://grafana.mlab-sandbox.measurementlab.net/d/deeyimsfzkwe8c/autojoin3a-site-overview?orgId=1&var-datasource=P9963F0EC00120018&var-org=garuda&var-machine=All

- - -
This change is [Reviewable](https://reviewable.io/reviews/m-lab/prometheus-support/1083)

### DIFF
--- a/config/federation/grafana/dashboards/Autojoin_SiteOverview.json
+++ b/config/federation/grafana/dashboards/Autojoin_SiteOverview.json
@@ -127,7 +127,7 @@
             "uid": "$datasource"
           },
           "editorMode": "code",
-          "expr": "(time() - node_boot_time_seconds{machine=~\"ndt-$machine.*\"}) / (60 * 60 * 24)",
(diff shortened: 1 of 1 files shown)


robertodauria (approved):
:lgtm:

> @robertodauria reviewed 1 of 1 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/m-lab/prometheus-support/1083#-:-OYkxq216W4uhVIGpjWY:bnfp4nl)* status: <span>:shipit:</span> complete! 1 of 1 approvals obtained

PR #1094, opened 2026-01-28
feat(prometheus): enable remote write receiver

robertodauria:
## Summary
- Enable the Prometheus remote write receiver by adding the `--web.enable-remote-write-receiver` flag
- This allows Prometheus to accept metrics via the remote write API

- - -
This change is [Reviewable](https://reviewable.io/reviews/m-lab/prometheus-support/1094)


bassosimone (approved):
This is fine because the deployment is such that nginx in front of it requires authentication as discussed right now with @robertodauria

PR #1095, opened 2026-03-03
Move webhook_config to the slack-alerts-ticket receiver for Alertmanager

nkinkade:
I noticed that lately when a production alert fires in the Slack #oti-alerts channel that a corresponding Github issue is not being created in the ops-tracker Github repository.

When I looked back at the last time that the metric `githubreceiver_created_issues_total` had a non-zero value, it was around January 17. Then looking at the latest release of this repository on January 28 (v2.65.0) that there were some changes I had made to the Alertmanager config. Upon closer inspection, I noticed that I had inadvertently moved the `webhook_config` section from the "slack-alerts-ticket" receiver to the newer "slack-fyis" receiver, which would explain why issues were no longer being created, since the `webhook_config` section is what configures Alertmanger to send a message to the alertmanager-github-receiver service.

This PR corrects that error, moving the `webhook_config` back to the right location.

Additionally, while testing the changes I noticed that the configmap-reload container in the Alertmanager pod was failing with:

```
error: Post "https://<user>:***@alertmanager.mlab-sandbox.measurementlab.net/-/reload": x509: certificate signed by unknown authority
```

So I also updated configmap-reload to the latest version with the hope that the updated container image might have an updated list of trusted CA certificates, which it appear to. I tested modifying the configmap manually, and configmap-reload noticed and successfully reloaded Alertamanager.

- - -
This change is [Reviewable](https://reviewable.io/reviews/m-lab/prometheus-support/1095)

### DIFF
--- a/k8s/prometheus-federation/deployments/alertmanager.yml
+++ b/k8s/prometheus-federation/deployments/alertmanager.yml
@@ -58,7 +58,7 @@ spec:
 
       # Check https://hub.docker.com/r/jimmidyson/configmap-reload/tags/ for the current
       # stable version.
-      - image: jimmidyson/configmap-reload:v0.7.1
+      - image: jimmidyson/configmap-reload:v0.9.0
         name: configmap-reload
         args: ["-webhook-url", "$(ALERTMGR_RELOAD_URL)", "-volume-dir", "/alertmanager-config"]
         env:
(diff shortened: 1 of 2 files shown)


k8s/prometheus-federation/deployments/alertmanager.yml
```diff
@@ -58,7 +58,7 @@ spec:
 
       # Check https://hub.docker.com/r/jimmidyson/configmap-reload/tags/ for the current
```
robertodauria: Nit: the comment here still references docker hub

robertodauria (approved):
LGTM. Thanks.

k8s/prometheus-federation/deployments/alertmanager.yml
nkinkade: Fixed.

PR #1096, opened 2026-05-20
Make IPv6 blackbox_exporter probes against a GCE instances instead of Linode

nkinkade:
The changes to the file IPV6_MONITORING.md in this PR pretty much explain what is happening here.  The basic idea is that instead of making IPv6 blackbox_exporter probes through a VM running in Linode, we're doing it through VMs that exist in GCE.

@robertodauria: you are already familiar with most all these changes, as you worked with me to create this change set with Claude Code (thanks!). Moving the probes to GCE VMs removes a fairly substantial amount of back-bending and complexity to make the Linode method work. That said, one small downside is that we'll now be running 3 GCE instances (one in each project) instead of a single VM in Linode. Then again, these GCE instances are of type e2-micro, which is not expensive.

- - -
This change is [Reviewable](https://reviewable.io/reviews/m-lab/prometheus-support/1096)

### DIFF
--- a/IPV6_MONITORING.md
+++ b/IPV6_MONITORING.md
@@ -1,82 +1,22 @@
 # IPv6 Monitoring
 
-Google Cloud Platform does not currently support IPv6 for most applications.
-Since Prometheus and the Prometheus blackbox\_exporter both run in GCP, we need
-another solution for IPv6 monitoring. The solution we decided to use is to run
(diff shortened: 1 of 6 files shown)


nkinkade:
Related PR: https://github.com/m-lab/terraform-support/pull/134

robertodauria (approved):

